Issue #6, opened 2015-06-26
Partners should have group based access permisions


xkmato:
Currently in Uganda one of the needs is put access restriction on some of the partners. Here a partners can monitor the same label but only for Ureporters in a specific group.

The whole point behind this is to have partners focus on messages from only specific regions(districts).

xkmato:
@ewheeler @rowanseymour @ericnewcomer so I am wondering if you guys have given this any though, and do you think there is any value or do we have to look at it in another perspective?

ewheeler:
so to clarify, in the Uganda case they might have a label `blue` and both partner orgs `Alice NGO` and `Bob NGO` both can see messages with label `blue`

BUT `Alice NGO` can only see messages that are labelled `blue` AND are from contacts in the group `Foo` -- whereas `Bob NGO` can only see messages that are labelled `blue` messages AND are from contacts in the group `Bar`

am i understanding this correctly? and to be more concrete these groups are typically grouping contacts by location/district? in other words `Alice NGO` is responsible for dealing with `blue` in `Arizona` while `Bob NGO` is responsible for dealing with `blue` in `Kansas`

xkmato:
Thanks for throwing clarity on this @ewheeler

rowanseymour:
This should be doable. We currently allow users to filter messages by group - my understanding is then that we would allow partner orgs to be associated with optional groups (in addition to labels) and then force filtering of messages by those groups.

rowanseymour:
Labelling logic can now be determined by group membership, and labels determine partner access.

PR #8, opened 2015-07-15
Adding group restriction to Partners

xkmato:
Reference to #6 
So before, I add anything to this. I just wanted to make sure that I am on the right direction.  
I can't seem to figure out how a contact is connected to group in casepro without making a request to rapidpro.

Please, if you get time, see below how I have attempted to do this and let me know if I am way off. No tests yet(Broken existing tests), I wanted to first understand the whole thing first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,5 @@ docs/_build/
 
 # PyBuilder
 target/
+
+celerybeat-schedule
--- a/casepro/cases/models.py
+++ b/casepro/cases/models.py
@@ -92,6 +92,10 @@ class Group(models.Model):
             else:
                 cls.create(org, group_names[group_uuid], group_uuid)
 
+    @classmethod
+    def get_for_contact(cls, org, contact_uuid):
+        return cls.filter(uuid__in=org.get_temba_client().get_contact(contact_uuid).groups)
+
     def as_json(self):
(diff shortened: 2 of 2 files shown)


nicpottier:
@rowanseymour can comment more when he gets back online next week, but the big limitation here is actually the RapidPro side. It is very expensive for us to do merges across contact groups on messages for orgs that have tens of millions of messages, so leaning on the API there probably isn't tenable.

We actually don't have a solid direction on how to implement this, the limiting factor is not casepro but rather the performance of the database at that scale.

rowanseymour:
The issue I'm seeing here is that we wouldn't be able to efficiently filter a list of cases (stored locally) by group access if we don't know the contacts' groups without hitting RapidPro. One solution there would be to store the group UUIDs on the case itself when it is created. The downside of that is that changes to the contacts groups on the RapidPro side wouldn't be reflected in CasePro, without the additional overhead of contact syncing which we've tried hard to avoid thus far.

If that's an acceptable compromise then it does lead us to the question of whether we should do something similar for messages and further reduce our reliance on the RapidPro API.

@xkmato let's discuss this before writing more code

xkmato:
Cool, let me know when you have time so we have chat about this. @rowanseymour

PR #18, opened 2016-03-11
Add has_labels flag to Message to speed up admin views

rowanseymour:
Admins see messages with all labels for their org but filtering over all labels to determine if a message belongs in the Inbox (i.e. it's labelled) or the Unlabelled folder, is slow. This adds a `has_labels` field which is maintained by a db trigger.

### DIFF
--- a/casepro/msgs/migrations/0027_message_has_labels.py
+++ b/casepro/msgs/migrations/0027_message_has_labels.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def populate_has_labels(apps, schema_editor):
+    Org = apps.get_model('orgs', 'Org')
+
+    for org in Org.objects.order_by('pk'):
+        print "Updating labels for org #%d..." % org.pk
+
+        for label in org.labels.all():
+            print " > Updating messages for label %s..." % label.name
+            label.messages.update(has_labels=True)
+
+
+class Migration(migrations.Migration):
+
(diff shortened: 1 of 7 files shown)


casepro/msgs/migrations/0027_message_has_labels.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ('msgs', '0026_auto_20160309_1337'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='message',
+            name='has_labels',
+            field=models.BooleanField(default=False),
```
rowanseymour: We're still small enough that I can get away with this

casepro/msgs/tests.py
```diff
@@ -252,6 +276,84 @@ def test_save(self):
 
         self.assertEqual(set(message.labels.all()), {feedback, important})
 
+    def test_search(self):
+        eric = self.create_contact(self.nyaruka, 'C-101', "Nic")
```
ericnewcomer: we are different people!

ericnewcomer:
I think this looks sane. How about following the sql migration pattern in a /sql directory as @nicpottier introduced in rapid?

nicpottier:
Ya, I had looked at this earlier and looked sane, similar comment to Eric's that as you add more triggers may become simpler to have the full state of those in single files going forward.

casepro/msgs/tests.py
rowanseymour: I forget that sometimes. Fixed!

PR #20, opened 2016-04-13
Advanced labelling rules

rowanseymour:
This PR adds groups and a field value test to the labelling criteria for each label. In the long run we should separate rules from labels, and allow rules to do things other than label such as archive and flag. This has been implemented with that in mind so there is a rules engine capable of doing those things, but for now we generate simple labelling rules from each label rather than treat rules as first class citizens.


casepro/backend/rapidpro.py
```diff
@@ -45,7 +47,7 @@ def local_kwargs(self, org, remote):
             'is_blocked': remote.blocked,
             'is_stub': False,
             'fields': fields,
-            SAVE_GROUPS_ATTR: groups,
+            Contact.SAVE_GROUPS_ATTR: groups,
```
rowanseymour: Just name-spacing these constants inside their classes

casepro/msgs/signals.py
```diff
@@ -5,17 +5,25 @@
...
 
-from .models import Message, SAVE_CONTACT_ATTR, SAVE_LABELS_ATTR
+from casepro.backend import get_backend
+
+from .models import Label, Message
+
+
+@receiver(pre_save, sender=Label)
+def update_label_uuid(sender, instance, **kwargs):
+    if instance.is_synced and not instance.uuid:
```
rowanseymour: Unsynced labels may have a UUID if they were previously a synced label

nicpottier:
Looks good, definitely agree with the separation of rules and labels there.

PR #21, opened 2016-05-11
Dynamic group support

rowanseymour:
* adds `is_dynamic` field to `Group`
* prevents trying to suspend contacts from or restore contacts to dynamic groups
* don't list dynamic groups as options for group suspension

Once this has been deployed and groups are re-synced on prod - then can push another change to unmark any dynamic groups as suspend groups.

### DIFF
--- a/casepro/contacts/models.py
+++ b/casepro/contacts/models.py
@@ -203,10 +203,10 @@ class Contact(models.Model):
                 raise ValueError("Can't suspend from groups as contact is already suspended from groups")
 
             cur_groups = list(self.groups.all())
-            suspend_group_pks = {g.pk for g in Group.get_suspend_from(self.org)}
+            suspend_groups = set(Group.get_suspend_from(self.org))
 
             for group in cur_groups:
-                if group.pk in suspend_group_pks:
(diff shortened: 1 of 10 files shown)


casepro/contacts/models.py
```diff
@@ -203,10 +215,11 @@ def suspend_groups(self):
     def restore_groups(self):
         with self.lock(self.org, self.uuid):
             for group in list(self.suspended_groups.all()):
-                self.groups.add(group)
-                self.suspended_groups.remove(group)
+                if not group.is_dynamic:
```
rowanseymour: We know contacts have been "suspended" from dynamic groups already which they shouldn't have

casepro/orgs_ext/forms.py
```diff
@@ -20,3 +26,49 @@ def __init__(self, *args, **kwargs):
     class Meta:
         model = Org
         fields = forms.ALL_FIELDS
+
+
+class OrgEditForm(forms.ModelForm):
```
rowanseymour: Have been keeping forms in their own forms module - just forgot about this one

Issue #23, opened 2016-05-12
Frontend code should use local model pks rather than backend ids


rowanseymour:
A holdover from the days when CasePro was a realtime view of data in RapidPro - some things are still referenced by their backend identifier in the front end code, e.g. messages, groups. These should all be switched to local primary key, as objects may be local only in some deployments.

rowanseymour:
Addressed for contacts in https://github.com/rapidpro/casepro/pull/33

rowanseymour:
Addressed for groups in https://github.com/rapidpro/casepro/pull/66, only messages remain

PR #25, opened 2016-05-12
Outgoing message improvements

rowanseymour:
- Splits the generic message sending endpoint into three more logical endpoints for case replies, bulk replies and forwards
- Records contacts and URNs on the outgoing message
- Fixes a bug on the forwarding modal dialog

### DIFF
--- a/casepro/backend/__init__.py
+++ b/casepro/backend/__init__.py
@@ -84,15 +84,12 @@ class BaseBackend(object):
         """
 
     @abstractmethod
-    def create_outgoing(self, org, text, contacts, urns):
+    def push_outgoing(self, org, outgoing):
         """
-        Creates an outgoing broadcast message
+        Pushes (i.e. sends) an outgoing broadcast message
 
         :param org: the org
-        :param text: the message text
-        :param contacts: the contact recipients
-        :param urns: the raw URN recipients
-        :return: tuple of the broadcast id and it's timestamp
+        :param outgoing: the outgoing message
         """
 
     @abstractmethod
--- a/casepro/msgs/views.py
+++ b/casepro/msgs/views.py
@@ -10,8 +10,6 @@ from smartmin.views import SmartCRUDL, SmartTemplateView
 from smartmin.views import SmartListView, SmartCreateView, SmartUpdateView, SmartDeleteView
 from temba_client.utils import parse_iso8601
 
-from casepro.cases.models import Case
-from casepro.contacts.models import Contact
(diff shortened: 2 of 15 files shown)


casepro/backend/__init__.py
```diff
@@ -84,15 +84,12 @@ def create_label(self, org, name):
         """
 
     @abstractmethod
-    def create_outgoing(self, org, text, contacts, urns):
+    def push_outgoing(self, org, outgoing):
```
rowanseymour: Previously the pattern for creating something that exists in both RapidPro and CasePro, was 

1. Use an API call to create it in RapidPro
2. Get the identifier used by RapidPro
3. Create locally with that identifier

But for things that aren't being pushed to the backend - this pattern makes more sense:

1. Create locally
2. Optionally push to backend and update with returned identifier

Another advantage of this is that the local outgoing message is still created even if the RapidPro API call fails.

casepro/msgs/views.py
```diff
@@ -245,29 +243,36 @@ def post(self, request, *args, **kwargs):
...
+        JSON endpoint for forwarding a message to a URN
+        """
+        @classmethod
+        def derive_url_pattern(cls, path, action):
+            return r'^message/forward/(?P<id>\d+)/$'
 
-            outgoing = Outgoing.create(request.org, request.user, activity, text, contacts, urns, case)
+        def post(self, request, *args, **kwargs):
+            text = request.POST['text']
+            message = Message.objects.get(org=request.org, backend_id=int(kwargs['id']))
```
rowanseymour: Currently the frontend uses the ids from RapidPro for messages but we should switch it to use local primary keys (https://github.com/rapidpro/casepro/issues/23). When we do then a view like this can be a simpler `SmartUpdateView`

static/coffee/modals.coffee
```diff
@@ -136,7 +136,7 @@ modals.controller('ComposeModalController', ['$scope', '$modalInstance', 'title'
 
     if $scope.form.$valid
       urn = {scheme: $scope.fields.urn.scheme, path: $scope.fields.urn.path, urn: ($scope.fields.urn.scheme + ':' + $scope.fields.urn.path)}
-      $modalInstance.close({text: $scope.fields.text, urn: urn})
+      $modalInstance.close({text: $scope.fields.text.val, urn: urn})
```
rowanseymour: Old bug... which some Angular unit tests might have picked up

nicpottier:
It's so beautiful!

PR #26, opened 2016-05-13
Further outgoing message improvements

rowanseymour:
* Added partner field to outgoing model which we'll need for any partner based view of sent messages
* Reworked case timeline so it doesn't rely on fetching messages from backend

### DIFF
--- a/casepro/backend/rapidpro.py
+++ b/casepro/backend/rapidpro.py
@@ -316,26 +316,26 @@ class RapidProBackend(BaseBackend):
             client.unlabel_messages(messages=[m.backend_id for m in messages], label=SYSTEM_LABEL_FLAGGED)
 
     def fetch_contact_messages(self, org, contact, created_after, created_before):
-        contact_json = contact.as_json()
-        label_map = {l.uuid: l for l in Label.get_all(org)}
+        """
+        Used to grab messages sent to the contact from RapidPro that we won't have in CasePro
+        """
+        contact_json = contact.as_json(full=False)
 
         # fetch remote messages for contact
         client = self._get_client(org, 2)
         remote_messages = client.get_messages(contact=contact.uuid, after=created_after, before=created_before).all()
 
         def remote_as_json(msg):
+            # should match schema of Outgoing.as_json()
             return {
-                'id': msg.id,
-                'contact': contact_json,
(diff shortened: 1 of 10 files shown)


casepro/backend/rapidpro.py
```diff
@@ -316,26 +316,26 @@ def unflag_messages(self, org, messages):
...
-                'flow': msg.type == 'flow',
-                'sender': None,
-                'broadcast': msg.broadcast
+                'direction': Outgoing.DIRECTION,
+                'case': None,
+                'sender': None
             }
 
-        return [remote_as_json(m) for m in remote_messages]
+        return [remote_as_json(m) for m in remote_messages if m.direction == 'out']
```
rowanseymour: For CasePro/RapidPro, we'll now rely on local messages entirely for incoming and only use this fetch to pick up additional sent messages from RapidPro

casepro/cases/tests.py
```diff
@@ -633,7 +621,7 @@ def test_timeline(self, mock_push_outgoing, mock_fetch_contact_messages):
         self.assertEqual(len(response.json['results']), 0)
 
         # messages won't have been fetched from the backend this time
-        mock_fetch_contact_messages.assert_not_called()
```
rowanseymour: This doesn't exist in Python 2.7

casepro/contacts/models.py
```diff
@@ -241,7 +241,7 @@ def release(self):
         self.is_active = False
         self.save(update_fields=('is_active',))
 
-    def as_json(self, full=False):
+    def as_json(self, full=True):
```
rowanseymour: Convention for all these `as_json(..)` methods is that by default they return complete representation - and `full=False` returns the shortened representation typically used when object is embedded in another object.

casepro/msgs/models.py
```diff
@@ -452,17 +452,16 @@ def create(cls, org, user, activity, text, contacts, urns, case=None):
 
     def as_json(self):
```
rowanseymour: Was previously trying to make JSON representation of `Outgoing` match that of `Message` but these models continue to diverge so it's not worth it.

ericnewcomer:
weee!

PR #28, opened 2016-05-16
Basic sent folder

rowanseymour:
There's definitely more could be added here such as a modal dialog to list out recipients (up to 1000 URNs and contacts), but this at least for now that lets admins see who is sending what and track down messages they see appearing in RapidPro.

screen shot 2016-05-16 at 14 40 04

### DIFF
--- a/casepro/settings_common.py
+++ b/casepro/settings_common.py
@@ -303,9 +303,13 @@ PERMISSIONS = {
 
     'orgs.org': ('create', 'update', 'list', 'edit', 'home', 'inbox'),
 
-    'msgs.label': ('create', 'update', 'list', 'unlabelled'),
+    'msgs.label': ('create', 'update', 'list'),
+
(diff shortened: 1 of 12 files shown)


casepro/settings_common.py
```diff
@@ -303,10 +303,14 @@
 
     'orgs.org': ('create', 'update', 'list', 'edit', 'home', 'inbox'),
 
-    'msgs.label': ('create', 'update', 'list', 'unlabelled'),
+    'msgs.label': ('create', 'update', 'list'),
+
+    'msgs.message': ('action', 'bulk_reply', 'forward', 'label', 'history', 'search', 'unlabelled'),
```
rowanseymour: Previously message wasn't a local model so we couldn't have these permissions so all message related views just used `orgs.org_inbox`. Now we can get a bit more specific.

nicpottier:
looks good.

PR #29, opened 2016-05-18
Even more Outgoing model changes

rowanseymour:
In the past, `Outgoing` was essentially just a record of a RapidPro broadcast along with which CasePro user sent it. As we start implementing reports around user activity, we need a more complete model of outgoing messages.

* Rather than using `Outgoing` to model a broadcast to multiple contacts, it is now a message sent to a single recipient. Existing instances which were sent to multiple contacts are split up.
* Adds `reply_to` field which references  the incoming message that is being replied to - essential for constructing incoming/response views of messages
* Removes empty messages and makes these impossible in future.

Also includes an improvement to the Sent folder view - each message now displays contact it was sent to and clicking on that contact filters the view by that contact (same as for incoming)

### DIFF
--- a/casepro/backend/rapidpro.py
+++ b/casepro/backend/rapidpro.py
@@ -5,6 +5,7 @@ import six
 from dash.utils import is_dict_equal
 from dash.utils.sync import BaseSyncer, sync_local_to_set, sync_local_to_changes
 from django.utils.timezone import now
+from itertools import chain
 
 from casepro.contacts.models import Contact, Group, Field
 from casepro.msgs.models import Label, Message, Outgoing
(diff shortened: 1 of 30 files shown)


casepro/backend/rapidpro.py
```diff
@@ -253,29 +253,39 @@ def create_label(self, org, name):
...
+                msg.save(update_fields=('backend_broadcast_id',))
 
     def add_to_group(self, org, contact, group):
         client = self._get_client(org, 1)
-        # TODO until we know which groups are dynamic, we should assume this call might fail
-        try:
-            client.add_contacts([contact.uuid], group_uuid=group.uuid)
-        except Exception:
-            pass
+        client.add_contacts([contact.uuid], group_uuid=group.uuid)
```
rowanseymour: We now do know which groups are dynamic and these methods won't be called for dynamic groups

casepro/msgs/migrations/0038_no_empty_messages.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def delete_empty_messages(apps, schema_editor):
+    Outgoing = apps.get_model('msgs', 'Outgoing')
+
+    empties = Outgoing.objects.filter(text=None)
```
rowanseymour: A long time ago the send message form didn't prevent sending empty messages - hence there's some of these in our db

PR #33, opened 2016-05-26
Basic contact page and contact displayer/identifier changes

rowanseymour:
- Switches frontend code to reference contacts by local id rather than backend UUID (which they might not have in some deployments)
- Allows deployments to switch between displaying contact names and anonymous identifiers using setting
- Adds very basic contact read page:

screen shot 2016-05-26 at 08 36 33

### DIFF
--- a/casepro/cases/context_processors.py
+++ b/casepro/cases/context_processors.py
@@ -4,23 +4,6 @@ import regex
 import time
 
 from django.conf import settings
-from urlparse import urlparse
-
-
-def contact_ext_url(request):
-    """
-    Context processor that adds values for constructing external contact URLs
-    """
-    # base URL for external contact links
-    if '://' in settings.SITE_API_HOST:
-        # host link is a full URL
-        components = urlparse(settings.SITE_API_HOST)
-        contact_ext_url_base = '%s://%s/contact/read/{}/' % (components.scheme, components.netloc)
-    else:
-        # host link is a host name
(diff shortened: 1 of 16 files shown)


casepro/cases/context_processors.py
```diff
@@ -4,23 +4,6 @@
 import time
 
 from django.conf import settings
-from urlparse import urlparse
-
-
-def contact_ext_url(request):
```
rowanseymour: Not used anymore - instead external contact URL is based on its own setting

casepro/contacts/models.py
```diff
@@ -245,12 +256,12 @@ def as_json(self, full=True):
         """
         Prepares a contact for JSON serialization
         """
-        result = {'uuid': self.uuid, 'name': self.name}
+        result = {'id': self.pk, 'name': self.get_display_name()}
```
nicpottier: Is it still possible for them to see the full backend uuid? Wasn't this something they rolled over before?

Just wondering whether there's value there for them cross referencing with RapidPro

nicpottier:
Looks good apart from that one Q

casepro/contacts/models.py
rowanseymour: They've only ever been able to see the first 6 chars of the UUID. We added links to the contact in rapidpro on the inbox for admins - with this change those links take a user to the casepro contact page, and from there they can jump to rapidpro by clicking "View External".

PR #35, opened 2016-05-26
Lock down export access for partner users

rowanseymour:
Added partner field to exports and only let partner users download exports that their partner org has created. Org admins can access all exports for their org.


nicpottier:
yep

PR #36, opened 2016-05-26
Allow some partners to have unrestricted label access

rowanseymour:
Some UPartners offices have created partner orgs which they use internally and explicitly have access to all labels. This isn't ideal from a display or performance point-of-view especially when number of labels is high.

This should also serve Momconnect usecase where partners aren't restricted by labels.

Also changes partner form to use choeckboxes rather than dropdown as this makes it easier to see which labels are selected.


nicpottier:
yarp

PR #40, opened 2016-05-30
Simplified Angular service methods and improved testing

rowanseymour:
* Replaces callback use with promises 
* Replaces use of jQuery's `$.param` with `$httpParamSerializer`
* Consistent and simpler service method names
* Test coverage of most service methods, all filter and util methods


rowanseymour:
screen shot 2016-05-30 at 15 28 25

karma/test-services.coffee
```diff
@@ -1,30 +1,410 @@
...
+        CaseService.fetchSingle(501).then((caseObj) ->
+          expect(caseObj).toEqual({id: 501, opened_on: utcdate(2016, 5, 17, 8, 49, 13, 698)})
+        )
+        $httpBackend.flush()
+      )
+    )
+
+    describe('reassign', () ->
+      it('posts to reassign endpoint', () ->
+        $httpBackend.expectPOST('/case/reassign/501/').respond('')
```
rowanseymour: Currently these tests aren't checking the post body because the requests use `FormData` and that's pretty awkward to mock or verify. In a subsequent PR I'll switch these to use urlencoded form submission which will make it easier to test here and make param handling simpler in general.

rowanseymour: On second thought - having them post JSON (the angular default) might actually be the simplest way to go.

ericnewcomer:
Nice, looks good!

PR #42, opened 2016-05-31
Fix transaction decorators on export creation views

rowanseymour:


nicpottier:
This would probably be useful in smartmin.. I actually think Smartmin's export suffers from the same problem. (it has a wait to retry as I recall)

PR #44, opened 2016-06-01
More frontend testing and cleanup

rowanseymour:
* all modal creating service methods now return promises rather than taking callbacks
 * server time now used for start point of item fetches
 * removed new item fetching code for cases - current implementation relies on browser time and so isn't reliable. Need to figure out mechanism for new item fetching that will work for messages and cases.
 * changed `BaseItemsController` to only load old items if active label actually changes - means item fetching is started by the `infinite-scroll` directive on the page rather than from controller initialisation
 * consistent use of parentheses for coffeescript function calls
 * more frontend test coverage

### DIFF
--- a/karma/helpers.coffee
+++ b/karma/helpers.coffee
@@ -1,4 +1,17 @@
-
+#
 # helper for creating UTC datetimes
+#
 utcdate = (y, m, d, h, min, sec, ms) ->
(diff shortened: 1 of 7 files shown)


rowanseymour:
screen shot 2016-06-01 at 15 46 50

PR #50, opened 2016-06-03
Misc tweaks and tests

rowanseymour:


casepro/backend/__init__.py
```diff
@@ -74,13 +74,12 @@ def pull_messages(self, org, modified_after, modified_before, as_handled=False,
         """
 
     @abstractmethod
-    def create_label(self, org, name):
+    def push_label(self, org, label):
```
rowanseymour: So it follows the pattern of backend methods taking existing objects and optionally doing something with them. Makes them easier to NOOP.

PR #52, opened 2016-06-07
Allow admin users to create other admin users and add password complexity checks

rowanseymour:
Reworks user editing UIs so that there 3 views:
1. Creating any kind of user from the org administration (accessible to org admins and superusers)
2. Creating a partner user from the partner dashboard (accessible to partner managers and org admins)
3. Updating any kind of user (accessible to all of above, tho partner managers can't change a user's partner)

Screenshot of 1 when logged in org admin:

screen shot 2016-06-07 at 10 29 05

Options reduced when logged in partner manager

screen shot 2016-06-07 at 11 11 36

### DIFF
--- a/casepro/cases/mixins.py
+++ b/casepro/cases/mixins.py
@@ -1,0 +1,16 @@
+from __future__ import unicode_literals
+
+from dash.orgs.views import OrgPermsMixin
+
+
+class PartnerPermsMixin(OrgPermsMixin):
+    """
+    Permissions mixin that requires users to have the view specific permission AND either be in the same partner org,
+    or not be attached to a partner.
+    """
+    def has_permission(self, request, *args, **kwargs):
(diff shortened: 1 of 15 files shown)


nicpottier:
Looks good, was password complexity something that actually came up?

I'm not a fan of the number, lower, upper sort of requirements, seems having a long password is way better and simpler for people to grock. But either way, looks good.

rowanseymour:
Someone admitted to having the password "password" and I figured we better put a stop to that

nicpottier:
12 characters or more!

rowanseymour:
I sorta agree... but there's also an argument for validation rules that users are more familiar with. Most people don't have 12 letter passwords. Either way, `is_password_complex` lives in Smartmin so would be nice to keep our agreed upon logic there. I like how it currently is that you can either include digit/upper/lower, or use a password with 12 or more characters. That seems to accommodate everyone.

PR #54, opened 2016-06-07
Change modelling of forwards so that there is one URN per Outgoing

rowanseymour:
Outgoing instances in CasePro used to model broadcasts in RapidPro. This makes forwards more like replies in that they model a single message, and is a preparation step for actually surfacing forwards somewhere in the UI - like the Replies tab on the partner pages.


nicpottier:
100% coverage! color me impressed!

rowanseymour:
Let's not get ahead of ourselves. It's only 99.929%.

PR #60, opened 2016-06-16
Labelling improvements

rowanseymour:
- Adds option to ignore single word messages
- Fixes bug on home page where labelling dropdown isn't populated until active label is changed
- Adds support for non-ascii keywords
- Changes keyword min length to 2
- Adds basic rule list page (not linked from anywhere for now, really just for debugging)

### DIFF
--- a/casepro/msgs/forms.py
+++ b/casepro/msgs/forms.py
@@ -5,6 +5,7 @@ from django.utils.translation import ugettext_lazy as _
 
 from casepro.contacts.models import Group
 from casepro.rules.forms import FieldTestField
+from casepro.rules.models import ContainsTest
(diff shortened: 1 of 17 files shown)


casepro/msgs/forms.py
```diff
@@ -5,26 +5,37 @@
...
 
 from .models import Label
 
 
 class LabelForm(forms.ModelForm):
     name = forms.CharField(label=_("Name"), max_length=128)
+
     description = forms.CharField(label=_("Description"), max_length=255, widget=forms.Textarea)
+
+    is_synced = forms.BooleanField(
```
rowanseymour: Moving this up so all the rule-related fields come last

PR #61, opened 2016-06-17
Don't let users remove labels they don't have access to

rowanseymour:
Users can apply individual labels to messages and cases, or open a labelling modal that let's them select multiple labels. This doesn't show labels that the user doesn't have access to, and these labels were then being removed when the modal form was submitted.


norkans7:
Looks good

PR #66, opened 2016-06-20
Updates ui-select dependency and use group pk rather than UUID in frontend

rowanseymour:


casepro/msgs/models.py
```diff
@@ -230,8 +230,8 @@ def search(cls, org, user, search):
 
         if contact_id:
             queryset = queryset.filter(contact__pk=contact_id)
-        if group_uuids:
-            queryset = queryset.filter(contact__groups__uuid__in=group_uuids).distinct()
+        if group_ids:
+            queryset = queryset.filter(contact__groups__pk__in=group_ids).distinct()
```
rowanseymour: Improves performance of searching by group

PR #67, opened 2016-06-21
Message Store and Identity Store integration for Junebug

imsickofmaps:
Including:

- [ ] Integration with Message Store
- [ ] Integration with Identity Store


imsickofmaps:
@rowanseymour this is not ready for review - just starting to follow the agreed process

miltontony:
@rowanseymour This PR is ready for review. 
It's the Junebug Backend implementation.
As agreed, in future we'll be pinging you on small individual PRs so as to avoid your first review being this huge PRs.

ewheeler:
only 3 lines shy of 100% test coverage! lets dream big ✨ 💯 ✨ 

https://coveralls.io/builds/7004208/source?filename=casepro%2Fbackend%2Fjunebug.py

casepro/backend/__init__.py
```diff
@@ -199,6 +199,15 @@ def fetch_contact_messages(self, org, contact, created_after, created_before):
                  Message.as_json() for incoming messages and Outgoing.as_json() for outgoing messages.
         """
 
+    @abstractmethod
+    def get_url_patterns(self):
+        """
+        Returns the list of URL patterns that should be registered for this
+        backend.
```
rowanseymour: Thus far we've had flake8 set with max-line-length=120, so no need to wrap to 79 (?)

I've tried to keep the code style to basically Django + 120 char wide lines.

casepro/backend/rapidpro.py
```diff
@@ -345,3 +345,7 @@ def remote_as_json(msg):
             }
 
         return [remote_as_json(m) for m in remote_messages if m.direction == 'out']
+
+    def get_url_patterns(self):
+        '''The Rapidpro backend doesn't have any urls to register.'''
```
rowanseymour: """ for docstrings

casepro/backend/__init__.py
rowanseymour: Another code convention that you guys either share or have figured out is that we tend to use '' for code strings and "" for display strings - when possible.

casepro/backend/junebug.py
```diff
@@ -0,0 +1,522 @@
...
+from ..msgs.models import Message
+from ..utils import uuid_to_int
+
+from dash.utils import is_dict_equal
+from dash.utils.sync import BaseSyncer, sync_local_to_changes
+
+from itertools import chain
+
+
+class IdentityStore(object):
```
rowanseymour: As I mentioned in dev list email, we would have added a URNs/addresses to the contact model in CasePro except that they're not needed in the UPartners case, which requires names and phone numbers to be hidden. Is that something that would be useful for you guys to add now, so that they can be kept in sync with this identity store and more easily surfaced in the UI at some point?

casepro/settings_common.py
```diff
@@ -56,6 +56,17 @@
 SITE_BACKEND = 'casepro.backend.NoopBackend'
 SITE_ANON_CONTACTS = False
 
+# junebug configuration
```
rowanseymour: This is fine for now but if we find ourselves needing lots of backend specific settings maybe we could move such things into `settings_rapidpro.py.dev` and `settings_junebug.py.dev`

rowanseymour:
Looks good. Just some questions re contact phone numbers, some code style issues, and a mere 0.083% more code coverage required.

ewheeler:
💯!!! nice one @rudigiesler

casepro/backend/junebug.py
rudigiesler: At the moment we fetch the URN from the identity store when we need it (when we send an outgoing message). It might lead to a small performance increase, but I don't see a strong argument for changing the way things work now.

If there's a good argument for syncing the URN when we sync the contact, and using that instead for sending outbound messages, then we can change things, but I don't think it's quite worth the effort at the moment, and as far as I know we have no need for surfacing it in the UI. (It might also introduce problems if a contact's URN gets changed in between syncs, and we send the message to the old URN).

PR #71, opened 2016-06-27
Tracking of daily outgoing message counts

rowanseymour:
Adds new statistics app with count models to track daily counts of outgoing replies, at org, partner and user levels.


casepro/cases/views.py
```diff
@@ -341,39 +341,6 @@ def get_summary(self, partner):
                 'cases_closed': Case.objects.filter(org=partner.org, assignee=partner).exclude(closed_on=None).count()
             }
 
-    class RepliesChart(OrgObjPermsMixin, SmartReadView):
```
rowanseymour: Moves to `statistics/views`. Can make chart views more re-usable when they're not tied to a CRUDL

casepro/statistics/models.py
```diff
@@ -0,0 +1,208 @@
...
+    @classmethod
+    def squash(cls):
+        """
+        Squashes counts so that there is a single count per unique field combination per day
+        """
+        unique_fields = list(cls.UNIQUE_FIELDS) + ['type', 'day']
+        unsquashed_values = cls.objects.filter(squashed=False).values(*unique_fields).distinct(*unique_fields)
+
+        for unsquashed in unsquashed_values:
+            with connection.cursor() as cursor:
```
rowanseymour: I appreciate these aren't the most readable dozen or so lines, but couldn't find another way of doing this that was both atomic, and generic, i.e. don't have duplicate for every different count model.

casepro/statistics/migrations/0001_initial.py
```diff
@@ -0,0 +1,60 @@
...
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('cases', '0037_partner_is_restricted'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DailyOrgCount',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('type', models.CharField(max_length=1)),
```
nicpottier: might want to give yourself more than a single char to grow..

casepro/statistics/models.py
nicpottier: So if this starts getting really big then this is going to get really slow, as it will essentially scan the whole table. On RapidPro we deal with this by keeping track of a high water mark of things we have squashed and only squash those types which have been inserted since that high water mark.

casepro/statistics/migrations/0001_initial.py
```diff
@@ -0,0 +1,60 @@
...
+        migrations.CreateModel(
+            name='DailyPartnerCount',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('type', models.CharField(max_length=1)),
+                ('day', models.DateField(help_text='The day this count is for')),
+                ('count', models.PositiveIntegerField(default=1)),
+                ('squashed', models.BooleanField(default=False)),
+                ('partner', models.ForeignKey(to='cases.Partner')),
+            ],
```
nicpottier: Playing devil's advocate, but are there advantages to splitting these out by org / user / partner vs having that encoded in the 'type' (say as a slug?)

Don't really know the answer here, just throwing it out there. @norkans7  went the route of essentially building slugs that represented what he wanted to keep count of from org_ids or user_ids etc.. Which leads to a single model that is maintained for counts etc.. 

Perhaps he can comment on the pros and cons there.

casepro/statistics/models.py
rowanseymour: So I was copying the RapidPro approach and then noticed that it's possible for a count to be inserted between the actual squashing and the lookup of the max id. I guess that's not the end of the world as it would just occasionally leave an unsquashed value.

nicpottier: Ya, there is an inherent race, but it isn't a big deal. Certainly better than trying to enumerate all every time. (I think)

casepro/statistics/migrations/0001_initial.py
rowanseymour: Ok so first attempt at this was keeping everything hierarchical - i.e. don't record explicit counts for orgs, but aggregate over all users. That started to get expensive, and complicated.

Second attempt was what you're describing. But that felt like it was getting messy because we already have a type field which will be used things like REPLIES vs FORWARDS vs INCOMING, but org vs partner vs user (lets call that "level") didn't easily fit into that. I ended up with all this type specific configuration for each level, and some methods only make sense for a single level.

In the end it felt simpler to pull the different "levels" into separate tables, but keep the thing being counted as a type field.

rowanseymour: I do like the idea of the slug field for determining which values should be squashed, but it seems like I'd still want foreign keys for org/partner/user for fast fetching of these counts. 

I also need to do label-based counts, so if I'm trying to do this with one table, I end up with with 4 foreign keys, 3 of which are null on each row, and then starts to look like something which should be normalised. Hmm..

casepro/statistics/models.py
```diff
@@ -0,0 +1,165 @@
...
+from django.core.cache import cache
+from django.db import models, connection
+from django.db.models import Sum
+from django.utils.functional import SimpleLazyObject
+from django.utils.translation import ugettext_lazy as _
+
+from casepro.cases.models import Partner
+
+
+class DailyCount(models.Model):
```
rowanseymour: So despite previous comments, I've gone with a single count model, and ditched all the foreign keys in favour of single scope field which has all that encoded. I was concerned about the performance of that, but there's an index on (item_type, scope, day) and it all seems fine.

The typical usage pattern as you'll see in the tests is to call a method to return a set of counts, and then call another method on that to aggregate it in the way you need, e.g.

```python
DailyCount.get_by_partner([partner1], DailyCount.TYPE_REPLIES).total()
DailyCount.get_by_org([org1, org2], DailyCount.TYPE_REPLIES).day_totals()
```

casepro/statistics/models.py
```diff
@@ -0,0 +1,165 @@
...
+            for encoded_scope, scope in six.iteritems(self.scopes):
+                total_by_scope[scope] = total_by_encoded_scope.get(encoded_scope, 0)
+
+            return total_by_scope
+
+    class Meta:
+        index_together = ('item_type', 'scope', 'day')
+
+
+def record_new_outgoing(outgoing):
```
norkans7: why not a classmethod or instance method?

norkans7:
Looks good

casepro/statistics/models.py
rowanseymour: Idea is that this records an outgoing message for the whole statistics module which might do more than just make instances of `DailyCount`

PR #74, opened 2016-06-29
Case activity notification changes

rowanseymour:
1. Adds migration to add users as watchers on existing cases
2. Case actions which automatically make the user a watcher now limited to:
 * Opening
 * Reopening
 * Adding a note
 * Replying

### DIFF
--- a/casepro/cases/migrations/0039_populate_case_watchers.py
+++ b/casepro/cases/migrations/0039_populate_case_watchers.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def populate_case_watchers(apps, schema_editor):
+    Case = apps.get_model('cases', 'Case')
+
+    cases = Case.objects.all()
+
+    for case in cases:
+        # get all users who have sent replies in this case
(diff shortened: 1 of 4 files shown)


casepro/cases/migrations/0039_populate_case_watchers.py
```diff
@@ -0,0 +1,37 @@
...
+        # get all users who have sent replies in this case
+        repliers = {o.created_by for o in case.outgoing_messages.all()}
+
+        # get all users who have acted on this case in a way that would now make them a watcher
+        actors = {a.created_by for a in case.actions.filter(action__in=('O', 'R', 'N'))}
+
+        watchers = repliers.union(actors)
+
+        for user in watchers:
+            case.watchers.add(user)
```
norkans7: no remove of other users here?

rowanseymour: Which other users? We haven't deployed the watcher stuff yet so cases shouldn't have any existing watchers. This is just to get existing cases setup as if the watcher stuff had been there previously.

norkans7: Ok, right

PR #83, opened 2016-07-08
Per-label notifications

rowanseymour:
Adds Watch/Unwatch buttons to label dashboard like what we have already on case pages:

 screen shot 2016-07-08 at 14 32 49

Adds a new notification model which allows us to check we're not sending multiple notifications for the same message, and also makes it very easy in future to build a Facebook-style list page of notifications. 

Have converted case notifications to use this new model and improved their emails so that they're more tailored to each different case action.


casepro/profiles/signals.py
```diff
@@ -0,0 +1,20 @@
...
+from django.db.models.signals import m2m_changed
+from django.dispatch import receiver
+
+from casepro.msgs.models import Label, Message
+
+from .models import Notification
+
+
+@receiver(m2m_changed, sender=Message.labels.through)
+def notify_label_watchers(sender, instance, action, reverse, model, pk_set, **kwargs):
```
nicpottier: Only madness leads down this path! My recommendation would be to make this explicit as opposed to automagic.

nicpottier:
Can you turn off notifications globally? Does it maybe make sense to have some kind of batching of these? Just imaging the flurry of emails that is going to be generated when someone batch-labels something.

rowanseymour:
When emails are being sent we could merge all message notifications for each user into a single email. Maybe let users try it as is and mention that as an option?

No such option to turn off notifications currently, but was thinking that users should at least have the option to switch off email notifications for their account. If notifications were exposed in the UI then it would make sense to let the emails be optional.

nicpottier:
I'm fine launching as is, but ya, can see both of those being needed with any real use.

casepro/profiles/signals.py
rowanseymour: Signalophobe. The signal approach was useful at first because `post_add` tells you which labels were actually added at the database level (and aren't duplicates), but now the `Notification` model itself prevents duplicate notifications. Have added methods on `Message` for labelling/unlabelling and notifications are now created explicitly from there.

PR #85, opened 2016-07-11
Reflect tab changes in url so that specific tabs can be linked to

rowanseymour:
For example, the URL `/org/home/#/labels` now goes directly to the labels tab on the org dashboard

### DIFF
--- a/karma/test-controllers.coffee
+++ b/karma/test-controllers.coffee
@@ -527,6 +527,59 @@ describe('controllers:', () ->
   )
 
   #=======================================================================
+  # Tests for HomeController
+  #=======================================================================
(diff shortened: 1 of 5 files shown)


nicpottier:
why the switch to numeric offsets for the tabs? seems like that is a lot more fragile in the future if you decide to reorder these or add new ones.

rowanseymour:
The slug strings are certainly better but ui-bootstrap insists on numerical indexes so at some point one has to translate between indexes and slugs.

nicpottier:
Fair enough, seems like that translation is pretty localized as well so not a big deal to change later if you have to.

PR #89, opened 2016-07-12
Exports of label and partner statistics

rowanseymour:
Adds export buttons to the Labels and Partners tabs..

screen shot 2016-07-12 at 16 00 16

Which bring up a date range selection modal

screen shot 2016-07-12 at 16 21 02

And lets users make Excel exports of daily counts of replies/incoming for each partner/label, e.g.

| Date             | Tea         | Coffee     | 
|-------------|----------|----------|
| 12-07-2016 | 12            | 34           |
| 13-07-2016 | 56           | 78           |

Includes some changes to make the export-related classes more generic so that they can be reused for new export types like this that aren't based on item searches the way existing export types are.


casepro/cases/tests.py
```diff
@@ -899,8 +897,7 @@ def test_watch_and_unwatch(self):
 
 
 class CaseExportCRUDLTest(BaseCasesTest):
-    @override_settings(SITE_ORGS_STORAGE_ROOT='test_orgs', CELERY_ALWAYS_EAGER=True,
```
rowanseymour: SITE_ORGS_STORAGE_ROOT='test_orgs' is now set on the base test class

templates/cases/email/case_export.html
```diff
@@ -1,13 +0,0 @@
-{% load tz %}
```
rowanseymour: All of these export type specific templates are replaced by _utils/email/export.html_ and _utils/email/export.txt_

templates/orgs_ext/org_home.haml
```diff
@@ -48,7 +48,11 @@
         .search-toolbar.clearfix
           - if perms.cases.partner_create or org_perms.cases.partner_create
             .pull-away
-              <a class="btn btn-default" href="{% url 'cases.partner_create' %}">{% trans "New..." %}</a>
+              .btn-group
+                %a.btn.btn-default{ href:"{% url 'cases.partner_create' %}" }
+                  - trans "New"
```
rowanseymour: Trying to be consistent and only use ellipses when a modal will appear

PR #91, opened 2016-07-14
Per label counts in inbox view

rowanseymour:
Adds a new `TotalCount` model in the statistics app so that item totals don't have to be calculated by summing over all daily counts. Uses this to provide inbox and archived totals for all labels.


casepro/cases/views.py
```diff
@@ -426,7 +426,9 @@ def get_context_data(self, **kwargs):
         user = self.request.user
         partner = user.get_partner(org)
 
-        labels = Label.get_all(org, user).order_by('name')
+        labels = list(Label.get_all(org, user).order_by('name'))
+        Label.bulk_cache_initialize(labels)
```
rowanseymour: Open to suggestions on a better pattern for putting extra data into JSON model representations etc - data which isn't always needed, and is best fetched in bulk.

templates/cases/inbox_base.haml
```diff
@@ -30,9 +30,8 @@
           %a.btn.btn-default{ ng-href:"/label/read/[[ activeLabel.id ]]/" }
             - trans "View"
 
-
     .row
-      .col-md-2
+      .col-md-3
```
rowanseymour: Widening the folders/labels sidebar to make room for the counts

ericnewcomer:
looks sane

PR #98, opened 2016-07-18
Display label names with underscores as spaces in inbox

rowanseymour:
To allow wrapping of long label names, e.g.

screen shot 2016-07-18 at 11 11 55
screen shot 2016-07-18 at 11 12 48


nicpottier:
yep

PR #101, opened 2016-07-19
Additional Partner information

imsickofmaps:
This is a smallish PR with extra information that will be used for other upcoming features. Partner orgs will have the following information associated with them: Time Zone, Description and Primary Contact.

(additional commits that haven't yet been landed from other PR's are currently present)


miltontony:
partner update 1
partner read 1
partner list

rowanseymour:
Some questions/feedback before I look at the code:
  1. What will `Partner.timezone` be used for? If this is for displaying of times in the UI then seems like it should be a property of user, as not all users belong to partner orgs, and not all partner orgs will belong to a single timezone.
  2. Primary contact on the partner page should be rendered as a link to the user's page
  3. I don't think timezone needs to appear on the partner list page, tho that depends on answer to question 1.
  4. Code coverage needs to get back to 100%.

gsvr:
@rowanseymour 
- We've removed the partner timezone (this was a communication problem - it's already on the organization so we don't need it), and rendered the primary contact as a link.
- Coverage is back to 100%.
- New screenshots are shown below.

partner list page
partner read page
partner update page

casepro/cases/forms.py
```diff
@@ -18,8 +17,9 @@ def __init__(self, *args, **kwargs):
         org = kwargs.pop('org')
         super(PartnerForm, self).__init__(*args, **kwargs)
 
+        self.fields['primary_contact'].queryset = org.get_users()
```
rowanseymour: This should be restricted to users within the partner org

casepro/cases/tests.py
```diff
@@ -1016,26 +1023,31 @@ def test_create(self):
...
+        response = self.url_post('unicef', url, {'name': "Helpers", 'description': "Helpers Description",
+                                                 'primary_contact': self.user1.pk, 'logo': None, 'is_restricted': True,
+                                                 'labels': [self.tea.pk]})
         self.assertEqual(response.status_code, 302)
 
         helpers = Partner.objects.get(name="Helpers")
         self.assertTrue(helpers.is_restricted)
         self.assertEqual(set(helpers.get_labels()), {self.tea})
+        self.assertEqual(helpers.description, "Helpers Description")
+        # self.assertEqual(helpers.primary_contact, self.user1)
```
rowanseymour: Not sure why this is disabled

rowanseymour:
Looks fine. Just think we should restrict partner primary contacts to being partner users. That means potentially unsetting a primary contact if their membership of the partner org changes. 

So probably just need a check for that in `User.update_role` and `User.remove_from_org`.

casepro/cases/forms.py
gsvr: @rowanseymour If that is the case, I think we should remove the `primary_contact` field from the partner `create` form. If you agree with that, are you happy if we use two different forms, like `PartnerCreateForm` and `PartnerUpdateForm`?

rowanseymour: @gsvr yeah that seems fine

gsvr:
@rowanseymour Please have another look.

casepro/cases/management/commands/bootstrap_dev.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import unicode_literals
```
rowanseymour: Is this supposed to be in this PR?

casepro/profiles/__init__.py
```diff
@@ -66,11 +66,24 @@ def _user_update_role(user, org, role, partner=None):
     elif role not in PARTNER_ROLES and partner:
         raise ValueError("Cannot specify a partner for role %s" % role)
 
-    # remove user from any partners for this org
-    user.partners.remove(*user.partners.filter(org=org))
+    # store the user's previous partner
```
rowanseymour: Seems like this could all be replaced by something like...

```python
remove_from = [p for p in user.partners_primary.filter(org=org) if p != partner]
user.partners_primary.remove(*remove_from)
```

rowanseymour:
Just couple of comments...

- Code coverage needs to get back to 100%
- `update_role` seems like it can be simplifed
- not sure if the bootstrap code is meant to be in this PR as not seeing any mention of it

casepro/cases/management/commands/bootstrap_dev.py
gsvr: This must be a merge problem. I'll refresh with the latest and see what happens.

gsvr: Not sure what happened. I'm removing it.

gsvr:
@rowanseymour I think everything is addressed now. Will you merge it if you're happy?

rowanseymour:
👍

PR #103, opened 2016-07-20
Change password hasher to cheaper one for test run speedup

rudigiesler:
There is a small change that we can make to the dev/testing settings file to improve test run time, from 379.215s to 278.501s on my local machine.


rowanseymour:
Neat. Merged as failing test is unrelated

PR #104, opened 2016-07-20
Set the password hasher to MD5 for tests only and not in settings.py.dev

rowanseymour:
@rudigiesler no sooner had I merged that last PR did I realise this changes it for my local dev instance even when not testing. Does change below work for you?


rudigiesler:
:+1: , this works for me.

PR #107, opened 2016-07-20
Ability to build case plugins for CasePro

miltontony:
casepro_data_plugin

Example of a dummy pod showing static data
https://github.com/praekelt/casepropods.dummy/


casepro/pods/base.py
```diff
@@ -0,0 +1,65 @@
+import json
+from confmodel import fields, Config as ConfmodelConfig
+from django.apps import AppConfig
+
+
+class PodConfig(ConfmodelConfig):
+    '''
```
rowanseymour: """ for docstrings

casepro/pods/tests/test_base.py
```diff
@@ -0,0 +1,22 @@
+import json
```
rowanseymour: We're using Python 2.x (as I assume you guys are? Never tested this thing in Python 3) so you'll want to start all your files with `from __future__ import unicode_literals`

casepro/pods/views.py
```diff
@@ -0,0 +1,36 @@
+import json
+from django.http import JsonResponse
+
+from casepro.pods import registry
+
+
+def read_pod_data(request, pk):
```
rowanseymour: Wouldn't this be clearer as `index`?

casepro/pods/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from casepro.pods.views import perform_pod_action, read_pod_data
+
+urlpatterns = (
+    url(r'read/(?P<pk>\d+)/$', read_pod_data, name='read_pod_data'),
```
rowanseymour: Might be better to reference pods by a configurable slug rather than a numerical index which is tied to the ordering of settings.POD, no? You guys are the ones using this so your call.

templates/cases/case_read.haml
```diff
@@ -151,9 +151,31 @@
                   [[ key ]]
                 : [[ val ]]
 
+        - for pod in pods
+            %div{
+              ng-controller:"={ pod.pod_type.controller }",
+              ng-init:'init && init(={ pod.config.index }, ={ case_id }, ={ pod.config_json })',
+              "={ pod.pod_type.directive }":"true"
+            }
```
rowanseymour: The existing directive cpContact is used as an element <cp-contact /> rather than attribute <div cp-contact.. >.  I don't mind which approach we use but would like to be consistent.

static/templates/pod.html
```diff
@@ -0,0 +1,12 @@
+<div class="panel panel-default">
```
rowanseymour: Can this live in _templates/partials/_ like other angular templates?

templates/cases/case_read.haml
justinvdm: +1 on using it as an element, will change.

static/templates/pod.html
rowanseymour: Note I've nicked the styling here for the details panel above it, which I was updating anyway in https://github.com/rapidpro/casepro/pull/110

justinvdm: Thanks for noting :) Now that we are adding action support, I think it is likely that we will change how this looks. In praekelt/casepro#48 (one of the smaller PRs for the next PR we plan on making later this sprint), in order to fit action buttons into the UI, we're changing it to look something like this (might still change a bit):

pod-with-actions

Thought it would be worth mentioning now to avoid inconsistencies with the contact panel.

templates/cases/case_read.haml
justinvdm: So after trying things out, I'm not sure how easily we'd be able to use directives as elements. The directive names depend on the context given to the template, and I can't seem to find a way of providing tags with names that depend on the context with hamlpy. Maybe it would be easier to stick with attributes?

rowanseymour: You can just add it as an html tag. The hamlpy library has a few limitations I'm going to look into at some point, so you'll notice that where a tag name contains a dash we have to fall back to html

justinvdm: Ah, thanks, didn't know that was possible, got it working as an html tag.

static/templates/pod.html
rowanseymour: I've updated https://github.com/rapidpro/casepro/pull/110 to be closer to that. You guys can always tweak it later.

justinvdm: @rowanseymour re your [templates question](https://github.com/rapidpro/casepro/pull/107#discussion_r71674335), we can do this. The tricky thing is getting haml templates working in karma tests for us to test our directive.

I've been trying out the karma haml preprocessors out there, none of them giving us what we need to be able to test:

  - [karma-ng-haml2js-preprocessor](https://github.com/codysims/karma-ng-haml2js-preprocessor): appears to look for a `haml` executable
  - [karma-haml-preprocessor](https://github.com/ondrasak/karma-haml-preprocessor): compiles to js strings that can be `eval`ed to raw html, but we're wanting to pipe its results to [karma-ng-html2js-preprocessor](https://github.com/karma-runner/karma-ng-html2js-preprocessor), which is expecting raw html

If we don't need to worry about using haml for this template and only need to keep the template in the `templates/partials/` directory, then this should be easy to change. Would this be alright?

rowanseymour: I guess the main argument for these being in templates/partials is that they can include django localization. Several of the modal partials do this.

If they're not using haml or any django tags... then maybe _static/templates_ is fine since that saves the app having to handle requests for them...

@ericnewcomer thoughts since you've done similar things in rapidpro?

casepro/pods/urls.py
rudigiesler: The reasoning behind using an index was that we wanted an auto generated thing, that you wouldn't be able to accidentally assign the same thing to two pods and have things break, and to just have less things to configure in general.

Whether this is a numerical index, or a UUID, or something else doesn't really matter.

The only issue I could think of is if you change the settings and restart the server, while someone still has the webpage open, it could break things until they refresh the page, but we would have the same issue with configurable slugs (if a pod is removed, we would still have issues).

It seemed that having less configuration to worry about was a win here, so we're happy with using the numerical index, although there doesn't seem to be a strong winner.

casepro/pods/base.py
rudigiesler: Resolved in https://github.com/rapidpro/casepro/pull/107/commits/327c985ee544d2b0d9330ad1a49e176b47d6cd3d

casepro/pods/tests/test_base.py
rudigiesler: Resolved in https://github.com/rapidpro/casepro/pull/107/commits/176e7372c818fbb24be1205a98871abc90d390c1

casepro/pods/views.py
rudigiesler: Resolved in https://github.com/rapidpro/casepro/pull/107/commits/5b2f959b9e2d7e96c59506f9476215ba2dffef40

casepro/pods/urls.py
rowanseymour: OK I was thinking that pod templates might need to manually reference these URLs but clearly that's not where you're going with these.

static/templates/pod.html
rowanseymour: I'm ok with leaving them where they are for now if you want to wrap this up. I think tho you have to change {{ to [[ for this particular template to work.

justinvdm: Thanks, will change to `[[`s. I see the interpolation symbols have been [configured](https://github.com/rapidpro/casepro/blob/master/static/coffee/app.coffee#L15-L16) to use `[[`s. For some reason, angular still seems to accept `{{` for interpolation, but makes sense to stick to `[[`s for consistency anyways.

rowanseymour: That is odd that it still works. Obviously {{ conflicts with Django hence the [[ for existing templates.

justinvdm:
@rowanseymour I think we've addressed all comments, ready for another review

static/coffee/services.coffee
```diff
@@ -549,3 +549,14 @@ services.factory('UtilsService', ['$window', '$uibModal', ($window, $uibModal) -
       return $uibModal.open({templateUrl: '/partials/modal_daterange.html', controller: 'DateRangeModalController', resolve: resolve}).result
 
 ])
+
+
+#=====================================================================
+# Pod API service
+#=====================================================================
+services.factory('PodApi', ['$window', '$http', ($window, $http) ->
```
rowanseymour: Sorry I didn't notice this before but can we call this `PodService` like the other services ?

PR #110, opened 2016-07-21
Render contact fields using labels

rowanseymour:
Also updates the styling of the contact details panel on the case page, to use same styling as that which will be used for pods provided by case plugins.

And ensures that date and numerical field values are formatted accordingly. So

screen shot 2016-07-21 at 14 57 05

becomes:

screen shot 2016-07-21 at 15 05 08

### DIFF
--- a/static/coffee/directives.coffee
+++ b/static/coffee/directives.coffee
@@ -28,3 +28,24 @@ directives.directive('cpContact', () ->
     ]      
   }
 )
+
+#----------------------------------------------------------------------------
+# A contact field value
+#----------------------------------------------------------------------------
+directives.directive('cpFieldvalue', () ->
+  return {
+    restrict: 'E',
+    scope: {contact: '=', field: '='},
+    template: '[[ value ]]',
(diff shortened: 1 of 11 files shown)


rowanseymour:
Updated contact details styling

screen shot 2016-07-21 at 13 56 48

static/coffee/directives.coffee
```diff
@@ -30,3 +28,27 @@ directives.directive('cpContact', () ->
     ]      
   }
 )
+
+#----------------------------------------------------------------------------
+# A contact field value
+#----------------------------------------------------------------------------
+directives.directive('cpFieldvalue', () ->
```
rowanseymour: @justinnvdm have added this to do formatting of date and numerical values, as we're currently just dumping ISO8601 timestamps on unfortunate users. Logic below should probably be shared in some way with pod values.

justinvdm: Thanks, noted. I think the plan for the base pod's UI is to expect all item values to be strings. If we start defining custom pods with values of other types, this will come in quite useful though.

PR #111, opened 2016-07-22
Translate language ISO codes into names when displaying

rowanseymour:
Also adds language to the contact details box on the case read page:

screen shot 2016-07-22 at 09 38 23

### DIFF
--- a/casepro/contacts/models.py
+++ b/casepro/contacts/models.py
@@ -9,6 +9,7 @@ from django.utils.translation import ugettext_lazy as _
 from django_redis import get_redis_connection
 
 from casepro.backend import get_backend
+from casepro.utils import get_language_name
 
 FIELD_LOCK_KEY = 'lock:field:%d:%s'
 GROUP_LOCK_KEY = 'lock:group:%d:%s'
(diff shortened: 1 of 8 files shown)


casepro/utils/__init__.py
```diff
@@ -164,3 +168,23 @@ def uuid_to_int(uuid):
     "Values from -2147483648 to 2147483647 are safe in all databases supported by Django"
     """
     return UUID(hex=uuid).int % (2147483647 + 1)
+
+
+def get_language_name(iso_code):
+    """
+    Gets a language name for a given ISO639-2 code.
```
rowanseymour: RapidPro uses ISO639-2 for language - if other deployments use a different code set then maybe do this translation in the backend abstraction.

casepro/contacts/models.py
```diff
@@ -269,6 +270,11 @@ def as_json(self, full=True):
         if full:
             result['fields'] = self.get_fields(visible=True)
 
+            if self.language:
```
norkans7: may be a method on the model, and reuse that in the views as well

casepro/contacts/views.py
```diff
@@ -44,6 +44,9 @@ def derive_fields(self):
 
             return fields
 
+        def get_language(self, obj):
+            return get_language_name(obj.language) if obj.language else None
```
norkans7: here use the model method

rowanseymour: This entire view is about to replaced in the next PR

norkans7:
Yep

PR #112, opened 2016-07-22
List a contact's cases on their read page

rowanseymour:
screen shot 2016-07-22 at 11 29 01

Note that the cases displayed are only the ones that the current user has access to.


nicpottier:
yarp

PR #115, opened 2016-07-25
Display contact status (blocked and/or stopped)

rowanseymour:
Adds `Contact.is_stopped` and displays both this and `Contact.is_blocked` on the contact and case read pages, if they are set. If they're not set, nothing is displayed so no confusion if these fields aren't being used.

On the contact read page:

screen shot 2016-07-25 at 14 49 15

and on the case page:

screen shot 2016-07-25 at 14 47 34

### DIFF
--- a/casepro/backend/rapidpro.py
+++ b/casepro/backend/rapidpro.py
@@ -45,13 +45,18 @@ class ContactSyncer(BaseSyncer):
             'name': remote.name,
             'language': remote.language,
             'is_blocked': remote.blocked,
+            'is_stopped': remote.stopped,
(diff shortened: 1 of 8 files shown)


norkans7:
why two parts migrations? isn't a boolean field with default enough?

rowanseymour:
Contact table on upartners.org is about ~2.5 million rows so can't add a new column with default because it will lock the table for about half an hour. Splitting into two migrations ensures that new contacts can be created with the correct default whilst the longer migration runs in the background.

norkans7:
Alright, looks good

PR #117, opened 2016-07-26
Add Frequently Asked Questions / Pre-approved response support

imsickofmaps:


casepro/msgs/models.py
```diff
@@ -132,6 +133,125 @@ def __str__(self):
 
 
 @python_2_unicode_compatible
+class Language(models.Model):
```
rowanseymour: What's the reason for modelling languages in the database rather than just storing codes and getting the actual language name from pycountry/iso639? 

We currently have a `Contact.language` field which is an ISO639-2 code and uses [iso639](https://pypi.python.org/pypi/iso639) to get the name for display. We can tweak that so that different encodings are supported (we're only using (ISO639-2 because that's what RapidPro uses), but I think we want `FAQ.language` and `Contact.language` to be modelled in the same way.

casepro/msgs/models.py
```diff
@@ -132,6 +133,125 @@ def __str__(self):
...
+
+    question = models.CharField(max_length=255)
+
+    answer = models.TextField()
+
+    language = models.ForeignKey(Language, verbose_name=('Language'), related_name='faqs', default=None)
+
+    parent = models.ForeignKey('self', null=True, blank=True, related_name='translations')
+
+    language = models.ForeignKey(Language, verbose_name=('Language'), related_name='faqs', default=None)
```
rowanseymour: These fields are defined twice

casepro/msgs/tasks.py
```diff
@@ -1,9 +1,23 @@
...
 from dash.orgs.tasks import org_task
 from datetime import timedelta
+from smartmin.csv_imports.models import ImportTask
+from .models import FAQ, Language, Label
+
+# python2 and python3 support
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
rowanseymour: I've not written a lot of Python 3 compatible code so might be wrong, but doesn't `io.StringIO` exist since Python 2.6?

casepro/msgs/tasks.py
```diff
@@ -89,3 +103,117 @@ def reply_export(export_id):
     logger.info("Starting replies export #%d..." % export_id)
 
     ReplyExport.objects.get(pk=export_id).do_export()
+
+
+def create_faq(org, question, answer, language, parent, labels=(), **kwargs):
```
rowanseymour: This can be moved to `FAQ.create(...)` like some of the other models

casepro/msgs/tasks.py
```diff
@@ -89,3 +103,117 @@ def reply_export(export_id):
...
+    """
+    A helper for creating FAQs since labels (many-to-many) needs to be added after initial creation
+    """
+    faq = FAQ.objects.create(org=org, question=question, answer=answer, language=language, parent=parent, **kwargs)
+    faq.labels.add(*labels)
+    return faq
+
+
+def get_or_create_lang(org, lang_code):
```
rowanseymour: Likewise, can be class method in `Language`

casepro/msgs/forms.py
```diff
@@ -79,3 +79,65 @@ def clean_keywords(self):
     class Meta:
         model = Label
         fields = ('name', 'description', 'is_synced', 'keywords', 'groups', 'field_test', 'ignore_single_words')
+
+
+class LanguageForm(forms.ModelForm):
+
+    code = forms.CharField(label=_("Code"), max_length=6,
```
rowanseymour: Having users enter ISO codes manually doesn't seem ideal. Would seem better to let them select languages by name using autocomplete if list is too long for a dropdown.

ewheeler: rapidpro does autocomplete:
https://github.com/rapidpro/rapidpro/blob/master/templates/orgs/org_languages.haml

casepro/msgs/migrations/0050_auto_20160712_1519.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('msgs', '0049_faq'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
rowanseymour: If possible please try to limit migrations within the same PR (e.g. in this case one could amend the original migration). I know this is tricky when code has already been deployed, but where we can do this it helps keep the number of migrations down. Also useful to ensure migrations have meaningful names tho I appreciate sometimes that's not worth it for migrations that do a lot of different things.

casepro/msgs/models.py
```diff
@@ -132,6 +133,125 @@ def __str__(self):
...
+        if name:
+            queryset = queryset.filter(name__icontains=name)
+
+        # Location filtering
+        if location:
+            queryset = queryset.filter(location__icontains=location)
+
+        return queryset.order_by('code')
+
+    def as_json(self):
```
rowanseymour: Follow pattern in other models where the arg `full` determines whether this should return the full representation, or just a short id/name representation used as a reference from another object.

casepro/msgs/tasks.py
```diff
@@ -89,3 +103,117 @@ def reply_export(export_id):
...
+            label = Label.objects.get(name__iexact=labelstring)  # iexact removes case sensitivity
+            labels.add(label)
+        except Exception as e:
+            task.log("Label %s does not exist" % labelstring)
+            raise e
+    return list(labels)
+
+
+@task(track_started=True)
+def faq_csv_import(org, task_id):  # pragma: no cover
```
rowanseymour: Is there a reason we can't use smartmin's CSV import stuff? Shouldn't require much more than making `FAQ` implement `SmartModel` and implementing `Smartmodel.create_instance`

casepro/msgs/tasks.py
```diff
@@ -89,3 +103,117 @@ def reply_export(export_id):
...
+    """
+    Gets a list of label objects from a comma-seperated string of the label codes, eg. "TB, aids"
+    """
+    labels = set()
+    labelstrings = labelstring.split(', ')
+    for labelstring in labelstrings:
+        labelstring = labelstring.strip()
+
+        try:
+            label = Label.objects.get(name__iexact=labelstring)  # iexact removes case sensitivity
```
rowanseymour: This needs to filter against org

casepro/msgs/tasks.py
```diff
@@ -89,3 +103,117 @@ def reply_export(export_id):
...
+        # Create the language
+        return Language.objects.create(org=org, code=lang_code)
+
+
+def get_labels(task, labelstring):
+    """
+    Gets a list of label objects from a comma-seperated string of the label codes, eg. "TB, aids"
+    """
+    labels = set()
+    labelstrings = labelstring.split(', ')
```
rowanseymour: Could use `casepro.utils.parse_csv`

casepro/msgs/faq_csv_example.csv
```diff
@@ -0,0 +1,4 @@
+Parent ID,Parent Language,Parent Question,Parent Answer,Labels,afr_ZA ID,afr_ZA Question,afr_ZA Answer,bla_ZA ID,bla_ZA Question,bla_ZA Answer
```
rowanseymour: Doesn't look like `Parent ID` is supported in imports?

rowanseymour: Also what is this file for?

casepro/msgs/forms.py
```diff
@@ -79,3 +79,65 @@ def clean_keywords(self):
...
+    class Meta:
+        model = Language
+        fields = ('code', 'name', 'location')
+
+
+class FaqForm(forms.ModelForm):
+
+    question = forms.CharField(label=_("Question"), max_length=140)
+    answer = forms.CharField(label=_("Answer"), max_length=140)
+    language = forms.ModelChoiceField(queryset=Language.objects.filter())
```
rowanseymour: Needs to be org-aware, i.e. only allow selecting items in the current org. Same for `parent` and `labels`. Can do this by passing org as form kwarg, and initialising fields in __init__

casepro/msgs/models.py
```diff
@@ -132,6 +133,125 @@ def __str__(self):
...
+    """
+    org = models.ForeignKey(Org, verbose_name=_("Organization"), related_name='faqs')
+
+    question = models.CharField(max_length=255)
+
+    answer = models.TextField()
+
+    language = models.ForeignKey(Language, verbose_name=('Language'), related_name='faqs', default=None)
+
+    parent = models.ForeignKey('self', null=True, blank=True, related_name='translations')
```
rowanseymour: Not sure about this parent/child relationship between translations. Seems like it would be simpler to have a single `FAQ` instance for each question, and then something like an HSTORE to hold all of that FAQ's different translations, no?

Also don't think `blank=True` does anything on `ForeignKey` fields

casepro/msgs/models.py
```diff
@@ -132,6 +133,125 @@ def __str__(self):
...
+    @classmethod
+    def search(cls, org, user, search):
+        """
+        Search for FAQs
+        """
+        language_id = search.get('language')
+        label_id = search.get('label')
+        text = search.get('text')
+
+        queryset = FAQ.objects.all()
```
rowanseymour: Needs to filter by org

rowanseymour:
Some general comments:
- Need to see some screenshots
- Several parts of the code aren't org-aware. Ideally everything should be unit tested in a way that ensures that users in one org can't access stuff in another org

casepro/msgs/views.py
```diff
@@ -117,32 +156,275 @@ def get(self, request, *args, **kwargs):
...
+
+        def derive_title(self):
+            return _("Import FAQs")
+
+        def post_save(self, task):
+            task.start(self.org)
+            return task
+
+
+class LanguageCRUDL(SmartCRUDL):
```
miltontony: @imsickofmaps @gsvr this is declared twice. First on `line: 182`

templates/frame.haml
```diff
@@ -79,12 +79,32 @@
...
                     %ul.dropdown-menu
+                      - if perms.orgs.org_home or org_perms.orgs.org_home
+                        %li
+                          %a{ href:"{% url 'orgs_ext.org_home' %}" }
+                            - trans "Organization"
+
+                      - if perms.msgs.label_list or org_perms.msgs.label_list
+                        %li
+                          %a{ href:"{% url 'msgs.label_list' %}" }
+                            - trans "Labels"
```
miltontony: Labels have been removed from the Administration Menu recently.

templates/frame.haml
```diff
@@ -79,12 +79,32 @@
...
+                - if request.user.is_superuser or org_perms.orgs.org_home or org_perms.msgs.label_list or org_perms.msgs.faq_list or org_perms.msgs.language_list or org_perms.cases.group_list or org_perms.profiles.profile_user_list
                   %li.dropdown
                     %a.dropdown-toggle{href:"#", data-toggle:"dropdown"}
                       - trans "Administration"
                       %b.caret
                     %ul.dropdown-menu
+                      - if perms.orgs.org_home or org_perms.orgs.org_home
+                        %li
+                          %a{ href:"{% url 'orgs_ext.org_home' %}" }
+                            - trans "Organization"
```
miltontony: Organization has moved from the Administration menu

templates/frame.haml
rowanseymour: See mailing list https://groups.google.com/forum/#!topic/casepro-dev/0J3YJp65p8w

miltontony: Thanks. Just making a note in the PR so that the team can rectify.

jpaidoussi:
This PR has been updated in the following ways to respond to the previous comments:

1. It has been brought more up to date with the current code
2. The migrations have been squashed into a single migration
3. The language model proposed has been removed and replaced with an iso639-2 field on FAQ to be more inline with the rest of the codebase
4. The other recommended code changes have been implemented

There are two parts of the feedback that have not been implemented:

1. Using an HSTORE field for FAQ translations rather than the parent-child model. This was not changed because we need to store three things for each translation - the language code, the translated question and the translated answer. An HSTORE field would only give us a key-string mapping.

2. We have not used Smartmin's CSV import mechanics for our CSV import. The reason here is because the import implemented by WCL has multiple FAQ entries per single CSV (a parent and all its potential translations) whereas the Smartmin import maps more directly to one row - one instance. 

This CSV format was recommended by Mike based on his experience in this field and the types of situations we may have to support.

Below are screenshots of the FAQ interface changes.

Viewing FAQS:
view faqs

Adding a FAQ:
add faq

Note: Adding a FAQ is not using an autocomplete field for language yet as it did not fit in the given timeframes. That can be added as an enhancement in the future.

Replying with a FAQ:
reply with faq

jpaidoussi:
@rowanseymour This PR is ready for review again.

casepro/cases/management/commands/bootstrap_dev.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import unicode_literals
```
rowanseymour: Don't think this file belongs in this PR

casepro/msgs/forms.py
```diff
@@ -79,3 +80,45 @@ def clean_keywords(self):
...
+        else:
+            parent = None
+
+        if parent is None and labels is None:
+            raise forms.ValidationError(_("Labels are required if no Parent is selected"))
+
+        if parent is not None:
+            labels = parent.labels.all()
+
+        # TODO 19: Update Translation FAQ labels when Parent FAQ labels are updated
```
rowanseymour: Would seem better to derive labels from parent rather than trying to keep parent and children in sync

casepro/msgs/forms.py
```diff
@@ -79,3 +80,45 @@ def clean_keywords(self):
...
+        if 'parent' in self.cleaned_data:
+            parent = self.cleaned_data['parent']
+        else:
+            parent = None
+
+        if parent is None and labels is None:
+            raise forms.ValidationError(_("Labels are required if no Parent is selected"))
+
+        if parent is not None:
+            labels = parent.labels.all()
```
rowanseymour: Would seem better to derive labels from parent rather than trying to keep parent and children in sync

casepro/msgs/forms.py
rowanseymour: As above

casepro/msgs/forms.py
```diff
@@ -79,3 +80,45 @@ def clean_keywords(self):
...
         fields = ('name', 'description', 'is_synced', 'keywords', 'groups', 'field_test', 'ignore_single_words')
+
+
+class FaqForm(forms.ModelForm):
+
+    question = forms.CharField(label=_("Question"), max_length=140)
+    answer = forms.CharField(label=_("Answer"), max_length=140)
+    language = forms.CharField(label=_("Language"), max_length=3)
+    # limit the parent choices to FAQs that have a ForeignKey parent that is None
+    parent = forms.ModelChoiceField(queryset=FAQ.objects.filter(parent=None), required=False)
```
rowanseymour: Needs to filter by org

casepro/msgs/forms.py
```diff
@@ -79,3 +80,45 @@ def clean_keywords(self):
...
+    labels = forms.ModelMultipleChoiceField(queryset=Label.objects.filter(), required=False, help_text=_(
```
rowanseymour: As above

casepro/msgs/models.py
```diff
@@ -171,6 +172,100 @@ def __str__(self):
...
+        return queryset.order_by('question')
+
+    @classmethod
+    def get_all(cls, org, label=None):
+        queryset = cls.objects.filter(org=org)
+
+        if label:
+            queryset = queryset.filter(labels=label)
+
+        return queryset.distinct()
```
rowanseymour: Shouldn't need distinct here

casepro/msgs/forms.py
jpaidoussi: This is handled here: https://github.com/praekelt/casepro/blob/EPIC-OPENHELPD-15-PR/casepro/msgs/views.py#L496-L502

casepro/msgs/forms.py
jpaidoussi: This is handled here as well: https://github.com/praekelt/casepro/blob/EPIC-OPENHELPD-15-PR/casepro/msgs/views.py#L496-L502

casepro/msgs/tasks.py
```diff
@@ -89,3 +100,80 @@ def reply_export(export_id):
     logger.info("Starting replies export #%d..." % export_id)
 
     ReplyExport.objects.get(pk=export_id).do_export()
+
+
+def get_labels(task, org, labelstring):
+    """
+    Gets a list of label objects from a comma-seperated string of the label codes, eg. "TB, aids"
```
rowanseymour: *separated

casepro/msgs/views.py
```diff
@@ -434,3 +454,178 @@ def post(self, request, *args, **kwargs):
...
+            context = super(FaqCRUDL.List, self).get_context_data(**kwargs)
+            # change the language code to full name for display only
+            for faq in context['object_list']:
+                faq.language = get_language_name(faq.language)
+            return context
+
+    class Create(OrgPermsMixin, SmartCreateView):
+        form_class = FaqForm
+
+        def customize_form_field(self, name, field):
```
rowanseymour: Better to do this in the form class since this overwrites whatever is there, and that's how it's done in forms like `LabelForm` etc.

casepro/msgs/views.py
```diff
@@ -434,3 +454,178 @@ def post(self, request, *args, **kwargs):
...
+        default_order = ('-parent', 'question')
+
+        def derive_queryset(self, **kwargs):
+            return FAQ.get_all(self.request.org)
+
+        def get_context_data(self, **kwargs):
+            context = super(FaqCRUDL.List, self).get_context_data(**kwargs)
+            # change the language code to full name for display only
+            for faq in context['object_list']:
+                faq.language = get_language_name(faq.language)
```
rowanseymour: Can use same pattern as `Contact.get_language()`

rowanseymour: If that's awkward to use in templates, we can switch both of those to return a non-model class, something like 

```python
class Language(object):
  def __init__(self, code):
    self.code = code
    self.name = get_language_name(code)

  def as_json():
    return {'code': self.language, 'name': self.name}

File: casepro/msgs/views.py [hunk 28]
```diff
@@ -434,3 +454,178 @@ def post(self, request, *args, **kwargs):
...
+        def save(self, obj):
+            data = self.form.cleaned_data
+            org = self.request.org
+            question = data['question']
+            answer = data['answer']
+            language = data['language']
+            parent = data['parent']
+            labels = data['labels']
+
+            faq = FAQ.objects.create(org=org, question=question, answer=answer, language=language, parent=parent)
```
rowanseymour: Should use `FAQ.create`

casepro/msgs/views.py
```diff
@@ -434,3 +454,178 @@ def post(self, request, *args, **kwargs):
...
+            labels = []
+            for label in self.object.labels.all():
+                labels.append(label.name)
+            context['labels'] = ', '.join(labels)
+            return context
+
+    class Update(OrgPermsMixin, SmartUpdateView):
+        form_class = FaqForm
+
+        def customize_form_field(self, name, field):
```
rowanseymour: As on `Create`, clearer to do this in one place on the form itself

casepro/msgs/views.py
```diff
@@ -434,3 +454,178 @@ def post(self, request, *args, **kwargs):
...
+        success_url = '@msgs.faq_list'
+
+        def derive_title(self):
+            return _("Import FAQs")
+
+        def post_save(self, task):
+            task.start(self.org)
+            return task
+
+    class Languages(OrgPermsMixin, SmartTemplateView):
```
rowanseymour: Not seeing where this class is used currently.. ?

Also not sure languages need to be paged since theres not going to that many per org. If we took out paging and added a search param this endpoint could be used for a language autocomplete.

jpaidoussi: It is used from the UI in the Reply Modal:
https://github.com/praekelt/casepro/blob/EPIC-OPENHELPD-15-PR/static/coffee/services.coffee#L193
https://github.com/praekelt/casepro/blob/EPIC-OPENHELPD-15-PR/static/coffee/modals.coffee#L68

We can remove the paging.

casepro/test.py
```diff
@@ -97,6 +109,11 @@ def create_label(self, org, uuid, name, description, keywords=(), **kwargs):
     def create_rule(self, org, tests, actions):
         return Rule.create(org, tests, actions)
 
+    def create_faq(self, org, question, answer, language, parent, labels=(), **kwargs):
+        faq = FAQ.objects.create(org=org, question=question, answer=answer, language=language, parent=parent, **kwargs)
```
rowanseymour: Can use `FAQ.create`

media/test_imports/faq_bad_import_labels.csv
```diff
@@ -0,0 +1,4 @@
+Parent ID,Parent Language,Parent Question,Parent Answer,Labels,afr ID,afr Question,afr Answer,bla ID,bla Question,bla Answer
```
rowanseymour: What is Parent ID column for ? Doesn't seem to be used

casepro/msgs/views.py
rowanseymour: Ah ok. Then I needs paging removed because that isn't paging through results

media/test_imports/faq_bad_import_labels.csv
jpaidoussi: I believe it is reserved for future use - specifically for allowing FAQ instances to be updated from the CSV upload.

static/coffee/services.coffee
```diff
@@ -157,11 +155,51 @@ services.factory('MessageService', ['$rootScope', '$http', '$httpParamSerializer
...
 #=====================================================================
+services.factory('FaqService', ['$rootScope', '$http', '$httpParamSerializer', ($rootScope, $http, $httpParamSerializer) ->
+  new class FaqService
+
+    #----------------------------------------------------------------------------
+    # Fetch FAQs to use in replies
+    #----------------------------------------------------------------------------
+    fetchFaqs: (search) ->
+      params = @_searchFaqsToParams(search)
+      return $http.get('/faq/search/?' + $httpParamSerializer(params)).then((response) ->
```
rowanseymour: As with languages - endpoint has paging but this doesn't page

rowanseymour:
Am going to use to mailing list to discuss some UI issues rather than cram it all in here. Appreciate the screenshots

miltontony:
@rowanseymour we're currently reworking UI based on discussions in the mailing list. Are you happy to land this PR as is so that the new PR is a delta rather than adding the new changes to this PR?

rowanseymour:
@miltontony we're running a production system and can't merge stuff into master which isn't ready to ship. Please continue in this PR or make a whole new PR if that is easier.

miltontony:
That's fine, when we submit the new PR, we can make it against this branch for easier reviewing. 

I'll get the coverage bumped.

rowanseymour:
It'll be easiest if there is one PR which has the finished code.

miltontony:
Cool, will ensure that this remains the final PR.

miltontony:
@rowanseymour ready for re-review. Please have a look at praekelt/casepro#142 for screenshots (and diff of changes).

static/coffee/services.coffee
miltontony: This was addressed in https://github.com/rapidpro/casepro/pull/117/commits/0cb3bf4cae45b6c1cb3a8d5bacf8e118501823b5

casepro/msgs/forms.py
miltontony: This has been addressed. When creating FAQ translations, you don't have to specify a label anymore. It automatically uses the labels of the parent.
https://cloud.githubusercontent.com/assets/9653693/19726469/557e7462-9b8b-11e6-9aee-5fb101a911c5.png

miltontony: Right.. we fixed that in the UI, but the code still made copies from the parent. I'm fixing that now and will ping you when it's done.

miltontony:
@rowanseymour ready for re-review. We've removed the copying of labels for translations and updated everything to always use the parent's label.

casepro/msgs/views.py
```diff
@@ -1,32 +1,52 @@
...
+    self.log("Queued import at %s" % now())
+    self.save(update_fields=['import_log'])
+
+    # trigger task
+    result = faq_csv_import.delay(org, self.pk)
+
+    self.task_id = result.task_id
+    self.save(update_fields=['task_id'])
+
+ImportTask.start = override_start
```
rowanseymour: Not loving this because it makes all imports FAQ imports... but I think it's ok for now since we don't currently import anything else

casepro/profiles/context_processors.py
```diff
@@ -16,3 +16,13 @@ def user(request):
         'user_is_admin': is_admin,
         'user_partner': partner
     }
+
+
+def user_must_reply_with_faq(request):
```
rowanseymour: This doesn't need to be a context processor - it's only used on one view

static/coffee/modals.coffee
```diff
@@ -58,10 +58,51 @@ modals.controller 'EditModalController', ['$scope', '$uibModalInstance', 'title'
 #=====================================================================
 # Reply to contacts modal
 #=====================================================================
-modals.controller('ReplyModalController', ['$scope', '$uibModalInstance', 'selection', 'maxLength', ($scope, $uibModalInstance, selection, maxLength) ->
+modals.controller('ReplyModalController', ['$scope', 'FaqService', '$uibModalInstance', '$controller', 'selection', 'maxLength', ($scope , FaqService, $uibModalInstance, $controller, selection, maxLength) ->
```
rowanseymour: Not seeing a screenshot of this modal

static/less/cases.less
```diff
@@ -62,14 +74,34 @@ footer {
...
+.input-clear .glyphicon {
+      pointer-events: all;
+    }
+
 /**
  * Smartmin tweaks - consistent layout for read page tables
  */
 .read-label {
-  width: 33%
+  width: 10%
```
rowanseymour: These tweaks are just for the FAQ read page right? If so then better to put them at the bottom of read template

miltontony:
@rowanseymour long overdue, but we've fixed those few small things that we left. Thanks!

casepro/settings_common.py
```diff
@@ -377,6 +382,8 @@
         'orgs.org_charts',
 
         'msgs.label_read',
+        'msgs.faq_search',
```
rowanseymour: Sorry just spotted this but I think Viewers need these permissions too

rowanseymour:
@miltontony need coverage back up to 100 and just one question about permissions

casepro/settings_common.py
miltontony: Updated for Viewers.

miltontony:
@rowanseymour coverage back at 💯 😉

rowanseymour:
Merged at last 🍾

miltontony:
🎉 🍻 🌟

PR #118, opened 2016-07-26
Improve appearance of login page

rowanseymour:
Before:

screen shot 2016-07-26 at 11 31 24

After:

screen shot 2016-07-26 at 11 32 04

### DIFF
--- a/templates/smartmin/users/login.haml
+++ b/templates/smartmin/users/login.haml
@@ -5,34 +5,34 @@
 - block login
 
 - block content
-  .row.login-form
-    .col-md-6.col-md-offset-3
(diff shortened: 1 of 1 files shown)


nicpottier:
Why not stick this in dash or smartmin? This is a shared template no?

rowanseymour:
Could do tho would need to make the "Email" vs "Username" thing configurable in some way, since CasePro is keeping email the same as username behind the scenes

nicpottier:
Hrmm.. ya, I guess changing the template is better than adding a new configuration option for all that stuff. looks good then.

Issue #120, opened 2016-08-03
Creating new administrator user redirects to user list JSON


rowanseymour:
This form used to redirect to a list of users - but that's now just JSON - so should redirect to the users tab of the org or partner org, depending on type of user

PR #121, opened 2016-08-03
Fix redirects when creating new users

rowanseymour:
* Creating a new admin user should redirect to users tab on org dashbaord
* Creating a new partner user should redirect to users tab on partner dashboard

Fixes https://github.com/rapidpro/casepro/issues/120


norkans7:
Looks good

PR #124, opened 2016-08-04
Don't use django-compressor with pod resources

rowanseymour:
Offline compression is currently broken so this just disables compression for pod resources. We can put it back in future, just need compression working again for now.

### DIFF
--- a/casepro/cases/tests.py
+++ b/casepro/cases/tests.py
@@ -373,9 +373,13 @@ class CaseTest(BaseCasesTest):
         self.assertEqual(case.access_level(self.user4), AccessLevel.none)  # user from different org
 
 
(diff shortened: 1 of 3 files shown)


casepro/cases/tests.py
```diff
@@ -373,10 +373,14 @@ def test_access_level(self):
         self.assertEqual(case.access_level(self.user4), AccessLevel.none)  # user from different org
 
 
+@modify_settings(INSTALLED_APPS={'append': 'casepro.pods.tests.utils.DummyPodPlugin'})
```
rowanseymour: Might as well test all methods here as if there's a pod loaded

justinvdm:
:+1: thanks for fixing, and sorry for breaking things. Getting compression working with dynamic context is probably going to be a bit tricky, missed that when I added this bit.

PR #125, opened 2016-08-04
Allow users to belong to different partners in different orgs

rowanseymour:
Replaces `Profile.partner` with `User.partners` / `Partner.users` many-to-many.

Users can still only be in one partner in a given org - but they can be in a partner in one org - and in another partner in another org.

Also moves `get_role` and `update_role` onto `User` class as `user.get_role(..)` makes more sense than `user.profile.get_role(..)` etc


ericnewcomer:
🎋

ewheeler:
nice tanabata tree!

PR #126, opened 2016-08-05
Add actions to pods

rudigiesler:
This allows the pods to specify a list of actions, which are displayed as buttons, which perform actions through callbacks to the pods.


casepro/pods/views.py
```diff
@@ -34,4 +38,15 @@ def perform_pod_action(request, index):
...
-    return JsonResponse(pod.perform_action(data))
+    case_id = data.get('case_id')
+    if case_id is None:
+        return JsonResponse(
+            {'reason': 'Request object needs to have a "case_id" field'}, status=400)
+
+    action_data = data.get('action', {})
+    success, payload = pod.perform_action(action_data.get('type'), action_data.get('payload', {}))
+    if success is True:
+        case = Case.objects.get(id=case_id)
```
rowanseymour: For security, filter by org too to make sure a user can never act on a case outside of their org

rowanseymour: Actually come to think of it, it should actually be checking `Case.access_level(user)` to make sure this user has correct access rights to this case.

Obviously this applies to anywhere requests can access or modify cases.

karma/test-services.coffee
```diff
@@ -630,16 +633,294 @@ describe('services:', () ->
...
+          })
+
+          $rootScope.$apply()
+
+          expect(document.querySelector('.modal-title').textContent)
+            .toMatch('Foo')
+
+          expect(document.querySelector('.modal-body').textContent)
+            .toMatch('Are you sure you want to do this?')
+        )
```
justinvdm: @rowanseymour I ended up modifying the approach we discussed a bit for modals. Is it supporting both the prompt-strings approach and the template url approach. What ended up differently though, was that instead of providing a prompt template url, one can provide a template url for the entire modal. The reason for this is that it gives us more flexibility for controlling how the modal looks. Thought I should bring this up since it is a slight change in what we discussed.

rudigiesler:
@rowanseymour Ready for review.

static/coffee/controllers.coffee
```diff
@@ -573,12 +588,12 @@ controllers.controller('CaseController', ['$scope', '$window', '$timeout', 'Case
   #----------------------------------------------------------------------------
 
   $scope.onWatch = () ->
-    UtilsService.confirmModal("Receive notifications for activity in this case?").then(() ->
+    UtilsService.confirmModal("Receive alerts for activity in this case?").then(() ->
```
rowanseymour: What's the reason for this change? Seems to me alerts and notifications are different things. 

I think I might have mentioned this on the dev list but I think at some point it would make sense to surface notifications in the UI. Should also ensure that pods have the ability to generate notifications and control how they are rendered as emails.

Also can we replace these calls with `ModalService.confirm(..)` calls so we're certain that new service can replicate the existing functionality, and we don't have two confirm modal methods.

justinvdm: Woops, this was an accidental replace from a search-and-replace, sorry! Will fix.

> Also can we replace these calls with ModalService.confirm(..) calls so we're certain that new service can replicate the existing functionality, and we don't have two confirm modal methods.

Can replace these calls here, but I'm a bit worried about replacing all confirm modals in this PR, get the feeling it will make this PR's scope grow quite a bit.

static/templates/modals/dummy-confirm.html
```diff
@@ -0,0 +1,12 @@
+<div class="modal-header">
```
rowanseymour: Do we need this file? Don't like having test files mixed up with real files and seems like one could test against static/templates/modals/confirm.html just as well

static/coffee/controllers.coffee
rowanseymour: Ok but you have to promise to change them all soon :)

static/templates/modals/dummy-confirm.html
justinvdm: We don't really, also wasn't too happy with mixing test files and real files. The reason for not testing against an existing template, is that the tests for a utility are using an unrelated piece of code, which breaks those tests' isolation a bit, and makes them dependent on an unrelated thing that may change or be removed at some point. How about we have a test templates dir for these (somehow didn't think of this earlier)?

static/coffee/controllers.coffee
justinvdm: /me nods

I'll give replacing them a try in this PR to get a sense of the level of effort required, will see from there.

static/templates/modals/dummy-confirm.html
rowanseymour: That works - and it can live in the karma folder

static/coffee/controllers.coffee
rowanseymour: I just took a peek and there's only ~15 calls I think - but some things you'll need to add/check:
- a danger param which sets the styling of the OK button to danger
- that prompt strings can include markup, e.g. `"Apply the label <strong>" + label.name + "</strong> to the selected messages?"` - or there's at least an easy way to do that without creating an entire modal template

justinvdm: Thanks for taking a look. I've created #127 for this, will make this change as a separate PR.

justinvdm:
@rowanseymour I think the review comments so far have been addressed

PR #128, opened 2016-08-11
add option to show full date and time on hover

miltontony:
case long date hover


miltontony:
@rudigiesler @justinvdm ready for review

miltontony:
@rowanseymour I used a similar implementation to what you guys use on RapidPro

rudigiesler:
Looks good to me, assuming that we want the american date format for the long dates.

rowanseymour:
I wonder if it wouldn't be better to use a tooltip like GMail / RapidPro, because for long messages - switching the date text like this will cause the text to re-arrange. Tooltips are pretty straightforward with uib-tooltip. Thoughts?

rowanseymour:
Also this behaviour for dates should be consistent throughout the app - maybe we can have a directive so that dates on the inbox views behave the same.

KaitCrawford:
@justinvdm I'm busy stumbling through the tests. Please take a look at my approach so long.

justinvdm:
@KaitCrawford approach looks good so far

KaitCrawford:
@miltontony @justinvdm @rowanseymour ready for re-review

KaitCrawford:
Tooltip as seen in the Inbox:
date_tooltip

karma/test-directives.coffee
```diff
@@ -60,6 +60,29 @@ describe('directives:', () ->
     )
   )
 
+  #=======================================================================
+  # Tests for tooltip
+  #=======================================================================
+  describe('cpTooltip', () ->
+
+    it('adds an element with a tooltip to the dom', () ->
+      $templateCache.put('/sitestatic/templates/tooltip.html')
```
justinvdm: Could you explain what this line is for?

karma/test-directives.coffee
```diff
@@ -60,6 +60,29 @@ describe('directives:', () ->
...
+      $scope = $rootScope.$new()
+
+      template = $compile('<cp-tooltip display-text="hello" tooltip-text="world" position="top" ></cp-tooltip>')
```
justinvdm: Should we add some content to the element here and test that we display this content?

justinvdm: Oh, sorry, that's what `display-text` is for, nevermind.

templates/partials/modal_messagehistory.haml
```diff
@@ -7,7 +7,7 @@
     .message-history
       .action.stackitem{ ng-repeat:"action in actions" }
         .action-time
-          [[ action.created_on | autodate ]]
+          <cp-tooltip display-text="[[ action.created_on | autodate]]" tooltip-text="[[ action.created_on | fulldate ]]" position="top-right"><cp-tooltip>
```
justinvdm: Wondering whether we want to be using the filters `autodate` and `fulldate` in the directive itself, or have them used when we give things to the directive (as we do at the moment). Both seem reasonable, just thought it was worth mentioning.

justinvdm:
Looks good :) just a minor comment.

karma/test-directives.coffee
KaitCrawford: uuuummm I don't know :blush: gone now

justinvdm:
:+1: from me when travis is happy.

templates/partials/modal_messagehistory.haml
KaitCrawford: a) that was my original plan but I couldn't get the directive to work when I tried to do that. I was probably doing it wrong since I don't really understand this stuff
b) currently this directive could be used to add tooltips to any text, it isn't restricted to dates. Which of course also means that it's not enforcing that dates always be displayed this way but it's just facilitating it.

rowanseymour: I'm not entirely convinced we get value from having a custom directive for tooltips which essentially implements uib-tooltip. What if we had a directive for displaying date values ? Something like:

```
<cp-date value="item.time" />
```

Which does the appropriate formatting and setup of the tooltip.

KaitCrawford:
@rowanseymour  The directive now applies the filters itself.
@justinvdm Please re-review. Specifically my use of angular.

templates/cases/case_read.haml
```diff
@@ -78,7 +78,7 @@
         .timeline{ ng-controller:"CaseTimelineController", ng-init:"init()" }
           .timeline-event.clearfix{ ng-repeat:"event in timeline | reverse", ng-class:'{ "timeline-action": (event.type == "A"), "timeline-incoming": (event.type == "I"), "timeline-outgoing": (event.type == "O") }' }
             .event-time
-              [[ event.time | autodate ]]
+              <cp-date-tooltip time=event.time position="top-right" />
```
justinvdm: Wondering if we should wrap `event.time` in quotes, something like `time="{{ event.time }}"`. Not sure if haml does that automatically?

justinvdm: If inspecting how the DOM for this looks suggests that haml handles this automatically, then happy with this.

KaitCrawford: I'm seeing `time="event.time"` in the DOM so it looks like haml handles it

justinvdm: Woops, sorry, meant `time="event.time"`. Sounds good, thanks for looking.

justinvdm:
👍

rowanseymour:
Sorry to be a pain on this but it still seems like this could be simpler. We want this used everywhere a date on a message or case is displayed - so that dates always have the same format and same behaviour when the user hovers over them.

Rather than how it is now where we still with a separate directive for the tooltip, i.e.

```
[[ event.time | autodate ]]
<cp-date-tooltip time=event.time position="top-right" />
```

Seems much simpler to have a directive that both 1) formats the date and 2) configures the tooltip, e.g.

```
<cp-date value="event.time" />
```

Internally the directive can call autodate to do the date formatting.

miltontony:
I suspect it might just be a naming issue. The current implementation is as follows:
```
<cp-date-tooltip time=event.time position="top-right" />
```
It's probably more accurate to call it `cp-date-with-tooltip` or simply `cp-date` but that's what the current implementation is (it's not just the tooltip).

rowanseymour:
Ah ok sorry my bad. Can we rename to `cp-date` and maybe get rid of the `position` attribute so it's always the same positioning? Or if there's a specific case where the positioning needs to be different than have a attribute called something like `tooltip-position` which defaults to something sensible.

static/coffee/directives.coffee
```diff
@@ -59,3 +59,15 @@ directives.directive('cpFieldvalue', () ->
...
+
+#=====================================================================
+# Tooltip directive
+# Shows 'displayText' with a tooltip at 'position' containing 'tooltipText'
+#=====================================================================
+directives.directive('cpDateTooltip', () ->
+  return {
+    restrict: 'E',
+    scope: {time: '=', position: '@'},
+    templateUrl: '/sitestatic/templates/tooltip.html',
```
rowanseymour: If you're ok with renaming directive to cp-date then let's also rename template to date.html

templates/cases/case_read.haml
rowanseymour: Is there a reason for omitting the quotes here? Could be just a bug in our HAML implementation that it works without them.

static/coffee/directives.coffee
miltontony: Agreed, let's rename this too.

KaitCrawford:
@justinvdm ready for re-review
@rowanseymour I've renamed things as you suggested, added the missing quotes and made the 'position' argument optional.

rowanseymour:
If we rename `position` to `tooltip-position` so it's clearer what that means, then I'm good.

KaitCrawford:
@rowanseymour rename done

PR #130, opened 2016-08-16
Assign user to case and automatic reassign

rudigiesler:
# Assign user to case
The idea behind this is that we not only want to assign a partner to a case, but also a specific user within that partner. This can later be tracked to see specific user performance, instead of general partner performance. For this PR, we're adding the ability to assign and reassign a case to a specific user within a partner, as well as showing that the specific user has been assigned in the correct places.

# Automatic reassign
When assigning a case to a partner, we don't want to assign it to them, and then them to do nothing with it without us knowing, we want it assigned back to us so the user isn't left without an answer. This PR adds a settings variable to set a specific amount of time (or `None` to disable the feature), and if a reassigned case is not closed within that preset timeframe, it is reassigned back to the original assignee.

@rowanseymour I think you've seen/been tagged in all of the PRs related to assigning a case to a user, thanks for your input there. For the automatic reassign, the relevant PRs are as follows:
https://github.com/praekelt/casepro/pull/74
https://github.com/praekelt/casepro/pull/80


casepro/cases/tasks.py
```diff
@@ -13,3 +14,40 @@ def case_export(export_id):
...
+        return
+
+    action = case.actions.filter(action=CaseAction.REASSIGN).latest('created_on')
+    if isinstance(action.created_by, SystemUser):
+        # The last time this case was reassigned was by the system, we should do nothing now, otherwise the
+        # assignment will just keep flipping back and forth
+        return
+
+    system_user = SystemUser.get_or_create()
+    note = _("This case's required response time as passed and therefor has been re-assigned")
```
ewheeler: "has passed"

casepro/cases/models.py
```diff
@@ -420,11 +465,21 @@ def access_level(self, user):
...
+
+        minutes_since_reassigned = (now() - self.last_reassigned_on).total_seconds() // 60
+        return minutes_since_reassigned > response_required_in
+
     def as_json(self, full=True):
         if full:
             return {
                 'id': self.pk,
                 'assignee': self.assignee.as_json(full=False),
+                'user_assignee': self.user_assignee.as_json(full=False) if self.user_assignee else None,
```
rowanseymour: `user_assignee` should be prefetched in `Case.search(...)` (and anywhere else we fetch more than one case)

casepro/cases/models.py
```diff
@@ -435,6 +490,7 @@ def as_json(self, full=True):
...
```
rowanseymour: Also need to prefetch this when searching messages since message representations include short case representations...

.. or just don't include `user_assignee` in the short form and just display the partner for simplicity when looking at messages?

rowanseymour:
The assigning-to-user stuff looks fine. Have some reservations about the approach for the auto-reassigning however, which I wish was in different PR.

- Currently the reassigned time is saved to the case and the un-assign task calculates the "expire" time from a django-setting. I think it would make more sense to record the "expire" time on the case - that way it's a lot easier to improve upon this feature in future by letting time limits be org configurable or case specific. 
- This only works for re-assigns, so cases have to be assigned to a partner and then re-assigned to another partner for it to take effect. It seems like we'd also want to support the use case where an org admin assigns a case to a partner, and wants to know if they don't respond.

casepro/cases/models.py
rudigiesler: Ah yes, the messages views. I'll add the template changes needed to show the assigned user to the various message views, it shouldn't be too big a change. Then I'll prefetch this as well.

rudigiesler: Resolved with https://github.com/rapidpro/casepro/pull/130/commits/c6b66b7e9c5351201946163a08aec0fd63765b30 and https://github.com/rapidpro/casepro/pull/130/commits/e50e46195efdad3bd522143d82d37d3912167a55

casepro/cases/models.py
rudigiesler: Resolved in https://github.com/rapidpro/casepro/pull/130/commits/e12c8c74c69305cecd39d2ad1910226cb439fa7f

casepro/cases/tasks.py
rudigiesler: Resolved with https://github.com/rapidpro/casepro/pull/130/commits/1aa764b463bc465d92f8dec8174af8e168c6094d

jpaidoussi:
@rowanseymour This is ready for review

rowanseymour:
@jpaidoussi can we get some screenshots of the UI changes. Also notice test coverage has dropped. Thanks

jpaidoussi:
Here is the UI for the added user assignee field:
user assignee

Here is an example of the auto reassign message:
auto reassign

rowanseymour:
Thanks. Some comments on the screenshots:

* On the re-assign dialog: would look neater with some spacing between those dropdowns. How about having separate labels for each dropdown - maybe "Assign to Partner" and "Specific User (Optional)"
* The default button on that dialog should be "Re-assign" rather than "Open" (seems it's always been like this but let's fix that while we're in here)
* On the re-assign note: "therefor" -> "therefore"

jpaidoussi:
@rowanseymour Updated:

reassign fixes

rowanseymour:
New screenshot looks good. However I've still got some issues with the automatic re-assignment stuff. Let's discuss on mailing list.

miltontony:
@rowanseymour following feedback from James, the UPartners guys won't be using this feature. We'll be conducting further workflow reviews on our side. So can we land this please?

rowanseymour:
@miltontony I've chatted with rest of team about this and we can't merge the automatic re-assignment functionality stuff as it is. The user assignment stuff is fine, and if you can find a way to separate that out, then we're happy to merge that.

As I've mentioned on the mailing list - the main problems are the fact this can't be used by org administrators, forces a re-assignment workflow, and the lack of transparency to users. UPartners can't/won't use the functionality in this PR, but enabling helpdesk administrators to find and act on stale cases is a problem we want to see solved for all deployments. We're happy to discuss further what that should look like.

rowanseymour:
Replaced by https://github.com/rapidpro/casepro/pull/169

PR #131, opened 2016-08-17
Fix pod rendering

justinvdm:
A recent change caused pod directives to no longer get their controller's scope anymore since the directive is no longer a child of the element to which the controller is applied.


miltontony:
Thanks @justinvdm. 
@rowanseymour review please

rowanseymour:
I think maybe I broke this when I ripped out the compression stuff? Anyways looks good.

PR #136, opened 2016-08-25
Bugs and tweaks

rowanseymour:
* Fixes "Select All" on inbox page which currently leads to Javascript error.
* Fixes label create form from redirecting to a JSON endpoint, and changes it and the partner and user create forms to all redirect to their model's read page.
* Tweaks partner form to use textarea for description like on label form.

### DIFF
--- a/casepro/cases/forms.py
+++ b/casepro/cases/forms.py
@@ -7,7 +7,9 @@ from casepro.msgs.models import Label
 from .models import Partner
 
 
-class PartnerUpdateForm(forms.ModelForm):
+class BasePartnerForm(forms.ModelForm):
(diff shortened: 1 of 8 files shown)


casepro/cases/forms.py
```diff
@@ -7,36 +7,37 @@
 from .models import Partner
 
 
-class PartnerUpdateForm(forms.ModelForm):
+class BasePartnerForm(forms.ModelForm):
+    description = forms.CharField(label=_("Description"), max_length=255, required=False, widget=forms.Textarea)
```
rowanseymour: Moving the common description/labels functionality into a base form class to avoid duplication

ericnewcomer:
🍡

PR #137, opened 2016-08-26
Add inline reply to message button

jpaidoussi:
The idea here was to improve end-user usability by adding a Reply button to the inline message actions between Open Case and Forward.

inline reply button

### DIFF
--- a/templates/partials/modal_reply.haml
+++ b/templates/partials/modal_reply.haml
@@ -5,7 +5,7 @@
     - trans "Reply"
   .modal-body
     %p
-      - trans "Send the following reply to each contact and archive each message. This will remove the messages from your inbox."
+      - trans "Send the following reply to the selected message(s). This will remove the message(s) from your inbox to Archived."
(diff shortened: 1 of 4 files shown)


templates/partials/modal_reply.haml
```diff
@@ -5,7 +5,7 @@
     - trans "Reply"
   .modal-body
     %p
-      - trans "Send the following reply to each contact and archive each message. This will remove the messages from your inbox."
+      - trans "Send the following reply to the selected message(s). This will remove the message(s) from your inbox to Archived."
```
rowanseymour: Can we just change this message completely when there is only one message? Something like..

"Send the following reply. This will archive this message removing it from your Inbox"

Think that we need to very clear to users that even if they have other messages selected - this won't reply to those.

jpaidoussi:
@rowanseymour Updated

rowanseymour:
👍

PR #138, opened 2016-08-26
Add time based metrics

jpaidoussi:
The requirement here was to add to minute based metrics:
- Average time taken from assignment to first response to a case by Partner
- Average time taken from assignment to closing a case by Partner

The Stats are displayed in Organisation > Partner Tab in the existing table as two additional columns:
average times


casepro/statistics/signals.py
```diff
@@ -52,3 +78,35 @@ def record_incoming_labelling(sender, instance, action, reverse, model, pk_set,
...
+@receiver(post_save, sender=CaseAction)
+def record_new_case_action(sender, instance, created, **kwargs):
+    org = instance.case.org
+    user = instance.created_by
+    partner = instance.case.assignee
+    case = instance.case
+
+    day = datetime_to_date(instance.created_on, instance.case.org)
+
+    if instance.action == CaseAction.CLOSE:
```
rowanseymour: What do we do with cases which are closed, re-opened and closed again?

casepro/statistics/signals.py
```diff
@@ -52,3 +78,35 @@ def record_incoming_labelling(sender, instance, action, reverse, model, pk_set,
...
+        # count the time to close on an org level
+        td = instance.created_on - case.opened_on
+        minutes_since_open = ceil(td.total_seconds() / 60)
```
rowanseymour: Still don't feel great about use of minutes for timing things - this adds an average of 30 seconds.

rowanseymour:
We seem to have a bit of screen real estate battle - both this PR and https://github.com/rapidpro/casepro/pull/139 add columns to the partners tab on the org dashboard (5 total) - which IMO is already pretty full at 3 columns.

Clearly not every piece of data we collect about a partner org can fit in that list view so we can we limit things to the partner read pages and exports?

casepro/statistics/signals.py
jpaidoussi: We are not counting those at this point. There was a bug in the code that has been fixed in the latest commits to make sure they are not counted.

casepro/statistics/migrations/0008_dailyminutetotalcount.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('statistics', '0007_populate_label_totals'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DailyMinuteTotalCount',
```
rowanseymour: So I know I reluctantly agreed to use minutes for timing on earlier PR but at that point I didn't know we were going to use them for timing message replies. Replies can happen within minutes, so it seems needlessly inaccurate to round everything up to the nearest minute. Changing this now to something like `DailyTiming` that uses seconds will be a lot easier than changing it later - so can we do that?

rowanseymour: This also needs a migration to calculate timings on historical data

jpaidoussi: @rowanseymour Please can you give us an indication of how many cases your live sites have so that we can plan the impact of the data migration?

rowanseymour: upartners.org (which hosts all of the UPartner orgs) currently has ~10,000 cases and ~150,000 replies

rowanseymour: If it's relevant there are also ~20,000,000 incoming messages (vast majority of messages are not replied to)

jpaidoussi: @rowanseymour For that many cases a data migration to calculate historic values is going to be slow and database intensive. Are you sure it is worth it for your use-case rather than just having totals on new cases?

rowanseymour: I don't think it would be that bad for only 10,000 cases. It's fine if it takes a few minutes, and better than having old cases show zero times.

miltontony:
Hey @rowanseymour, we've changed the implementation to use Seconds instead of minutes. We've also added a management command to calculate the stats for existing data. This PR is ready for review again.

miltontony:
@rowanseymour we're busy updating #139 (and this PR) to use data migrations instead of the management command. Can you please confirm that you have reviewed this PR aside from the data migration change?

rowanseymour:
@miltontony I asked in the slack channel if I should hold off re-reviewing this until the migration was fixed, and #139 was merged since I think these two have some conflicts. That work for you?

Once #139 is merged and this is updated I'd like to do some quick performance testing of the dashboard views on a copy of our prod database.

miltontony:
@rowanseymour I didn't see the ping in slack since you didn't mention me. There will likely be a conflict in the name and number of the migration. Apart from that, please review this PR - there are things in this PR that are not related to the other PR. For one, the changes from minutes to seconds is specific to this PR and we'd like to know if you're happy with the changes here. There were some other small things that were mentioned in previous reviews that we fixed.

casepro/statistics/signals.py
```diff
@@ -38,6 +41,29 @@ def record_new_outgoing(sender, instance, created, **kwargs):
         if instance.partner:
             DailyCount.record_item(day, DailyCount.TYPE_REPLIES, partner)
 
+        if case:
+            # count the very first response on an org level
+            if instance == case.outgoing_messages.earliest('created_on'):
```
rowanseymour: This is recording the time between a case being opened and the first reply sent by a user. Aren't we actually interested in the time between an incoming message being received and the first reply? That's certainly what I assumed from the header "Average time to reply" on the org dashboard.

If partners are opening their own cases, and immediately sending a reply - then this number will make it look like the partner is very responsive - but the incoming messages may have been sitting there for a long time with no reply.

casepro/utils/__init__.py
```diff
@@ -185,3 +185,23 @@ def get_language_name(iso_code):
         LANGUAGES_BY_CODE[iso_code] = lang
 
     return LANGUAGES_BY_CODE[iso_code]
+
+
+def humanise_seconds(seconds):
+    days, secs = divmod(seconds, 86500)
```
rowanseymour: This should be 86400, no? Anyways we can leverage Django's own time formatting classes here, and that's how we already format durations on the replies tab of the partner dashboard, e.g.

```python
def humanize_seconds(seconds):
    now = timezone.now()
    return timeuntil(now + timedelta(seconds=seconds), now)
```

`humanise` should also be `humanize` to match the US spelling used elsewhere

casepro/cases/views.py
```diff
@@ -413,15 +413,27 @@ def render_as_json(self, partners, with_activity):
                                                        *month_range(0)).scope_totals()
                 last_month = DailyCount.get_by_partner(partners, DailyCount.TYPE_REPLIES,
                                                        *month_range(-1)).scope_totals()
+                average_replied = DailySecondTotalCount.get_by_partner(partners,
+                                                                       DailySecondTotalCount.TYPE_TILL_REPLIED)
+                average_replied = average_replied.scope_averages()
```
rowanseymour: This should be within a given time window rather than an overall average, e.g. average response time in last month. I think it's more meaningful and it'll be more performant.

casepro/cases/views.py
```diff
@@ -413,15 +413,27 @@ def render_as_json(self, partners, with_activity):
...
-                        'this_month': this_month.get(partner, 0),
-                        'last_month': last_month.get(partner, 0),
-                        'total': total.get(partner, 0)
-                    }
+                    obj.update({
+                        'replies': {
+                            'this_month': this_month.get(partner, 0),
+                            'last_month': last_month.get(partner, 0),
+                            'total': total.get(partner, 0),
+                            'average': humanise_seconds(average_replied.get(partner, 0))
```
rowanseymour: 'average' is too vague here

rowanseymour:
@miltontony have re-reviewed

casepro/utils/__init__.py
jpaidoussi: timesince and timeuntil from Django do not have second resolution, only minute.

rowanseymour: I think that's ok. My main objection to minutes for storing times was accumulating inaccuracy. For display, it's unlikely anyone's averages will actually be less than 1 minute. 

The nice thing about Django's timesince is that it's localised, and at some point we're going to have to support proper localization of the UI.

casepro/statistics/signals.py
miltontony: @rowanseymour I've checked the requirements on our side - we specifically need to now the average time from when a case was opened. For our use case, the national team will be triaging the messages and it's only when a case is opened and assigned to a provincial officer (partner) do we need to look at the responsiveness. Of particular interest we need to understand how responsive the provincial team is at dealing with cases.

rowanseymour: There's two workflows we need to support here - first is:

1. Helpdesk admin assigns a case to a partner
2. Partner replies to contact
3. Partner responsiveness is time between being assigned and reply sent

Second is:

1. Partner opens case from messages they have label access to
2. Partner replies to contact (likely immediately)
3. Partner responsiveness is time between being message being received and reply sent

UPartner deployments use both workflows. For the second workflow, only measuring time between assignment and replying is at best meaningless and at worst misleading.

I gather it's not an immediate requirement for you guys, but time between a message being received, and a reply being being sent is meaningful for any deployment. It tells about the quality of service from the point of view of the contact, so I think that's something that should be recorded regardless of what else is recorded.

Where cases are being assigned to partners, then it's also clearly useful for any deployment to know how quickly partner orgs respond. So maybe that's something we record separately as "time between assign/re-assign and first reply from the partner". It would not apply when the partner self-assigns the case.

That make sense?

miltontony: The first scenario is definitely one that caters for our case. 
The second scenario does warrant consideration, but I don't think looking at averages alone will solve the oversight problems. If a partner reports a very low average for initial response, but then has a horrible case-closing record, that would show a potential abuse of the system by immediately responding. But it might also just speak to the different workflows. If partners always respond to cases with "Thank you for your message. We're looking into it" then the _average time to respond_ becomes a useless indicator.

I'm happy to add the second scenario to our TODO list, but I don't see it as a blocker at the moment. Also, I've been having some conversations with our team for looking at different ways of visualising this information, because the current table is not really ideal for this. You'd much rather have a graph that shows historical data than just a "Last month/This month" number. We'll add this to our TODO list as well.

We can definitely spend more time hashing out different use cases and potential abuses, but we

miltontony: Cool, that sounds like a reasonable compromise. I'll start on the PR for ignoring self-assigned cases and the rename.
re mege conflicts - I pushed the fix for that this morning.

miltontony:
@rowanseymour this is ready for re-review

PR #139, opened 2016-08-26
Add case counts to partners & users

smn:
This adds counters for partners & users, both in the UI and the exports.


casepro/statistics/signals.py
```diff
@@ -52,3 +53,22 @@ def record_incoming_labelling(sender, instance, action, reverse, model, pk_set,
...
+    org = instance.case.org
+    user = instance.created_by
+    partner = instance.case.assignee
+
+    day = datetime_to_date(instance.created_on, instance.case.org)
+    if instance.action == CaseAction.OPEN:
+        DailyCount.record_item(day, DailyCount.TYPE_CASE_OPENED, org, user)
+        DailyCount.record_item(day, DailyCount.TYPE_CASE_OPENED, partner)
+
+    elif instance.action == CaseAction.CLOSE:
```
rowanseymour: Maybe we only count if this is the first time this case has been closed? Otherwise users can up their stats by closing/reopening cases.

static/coffee/controllers.coffee
```diff
@@ -485,7 +485,7 @@ controllers.controller('HomeController', ['$scope', '$controller', 'LabelService
         $scope.labels = labels
       )
     else if tab == 'users'
-      UserService.fetchNonPartner(true).then((users) ->
+      UserService.fetchAll(true).then((users) ->
```
rowanseymour: This changes the org users tab to include all users across all partner orgs. 

There's certainly an argument for including them on that tab but I think it would be a checkbox that the user can select, and we'd want to add new fields to show which partner each user is in.

jpaidoussi: @rowanseymour We have now added a select/dropdown for filtering and the partner name is already included in the display:

org users

select filter

miltontony:
@rowanseymour ping from here - not sure what's outstanding on this PR. Can you please review?

rowanseymour:
@miltontony see https://github.com/rapidpro/casepro/pull/138#issuecomment-243191085

static/coffee/controllers.coffee
rowanseymour: Can we change this to a control which indicates which users are being displayed without the user having to click on it?

It isn't obvious that a user should click something that just says "Select" to change the view of users.

templates/orgs_ext/org_home.haml
```diff
@@ -74,6 +74,12 @@
               - trans "Replies last month"
             %th
               - trans "Total replies"
+            %th
+              - trans "Cases Opened this month"
```
rowanseymour: Other column headers aren't using title case

rowanseymour:
Needs migration to calculate case totals on historical data

miltontony:
Hey @rowanseymour, we've added a management command to calculate the stats for old cases. We've also addressed the other comments raised.

rowanseymour:
@miltontony why a management command and not a migration like discussed?

miltontony:
@rowanseymour given the potential slowness of the migration, there was a concern on idempotency of the run. If the migration were to fail midway, we wanted to provide a way to re-run the migration whilst maintaining the integrity of the data already calculated. We've thus allowed you to specify a `case-id` to begin at incase migration ran up to a certain point.

casepro/statistics/models.py
```diff
@@ -249,22 +252,73 @@ def render_book(self, book):
                 self.write_row(sheet, row, [day] + totals)
                 row += 1
 
+        elif self.type == self.TYPE_USER:
+            replies_sheet = book.add_sheet(six.text_type(_("Replies Sent")))
+            cases_opened_sheet = book.add_sheet(six.text_type(_("Cases Opened")))
+            cases_closed_sheet = book.add_sheet(six.text_type(_("Cases Closed")))
+
+            users = self.org.get_org_users().order_by('pk')
```
rowanseymour: Name would be a friendlier ordering. Also I think prefetching profiles will speed things up a little.

rowanseymour: Also if the user is looking at non-partner users, this export should reflect that.

templates/orgs_ext/org_home.haml
```diff
@@ -141,10 +153,27 @@
...
-          .{ style:"margin-top: 7px" }
-            Total of <strong>[[ users.length ]]</strong> users
+              .btn-group
+                %a.btn.btn-default{ href:"{% url 'profiles.user_create' %}" }
+                  - trans "New"
+                %button.btn.btn-default{ ng-click:"onExportUserStats()" }
+                  - trans "Export Users..."
+
+          .selection-controls.pull-back
+            .btn-group{ uib-dropdown:"true" }
```
rowanseymour: As I commented above, this would be better as a checkbox like "Include partner users" so that users can see what they are looking at without clicking a dropdown.

I think it should also default to false because some of our orgs are quite large, and fetching activity data for all users is an expensive query.

casepro/statistics/management/commands/calculate_case_open_close_totals.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import unicode_literals
+
+from django.core.management.base import BaseCommand
+from casepro.cases.models import Case, CaseAction
+from casepro.statistics.models import datetime_to_date, DailyCount
+
+
+def calculate_totals_for_cases(case_id, progress_callback=None):
```
rowanseymour: This should be a migration - will explain more in other thread.

rowanseymour:
@miltontony I tested the management command on a copy of our prod database and it took ~2 mins. Idempotency isn't an issue because migrations are run in transactions. If you added some prefetching you could probably make it even faster, but ~2 mins is not a problem.

Please change this to a migration (same for https://github.com/rapidpro/casepro/pull/138) we don't want to clutter the codebase with management commands everytime we need to do a data migration.

casepro/statistics/management/commands/calculate_case_open_close_totals.py
```diff
@@ -0,0 +1,59 @@
...
+    qs = qs.order_by('id')
+    for case in qs:
+        open_action = case.actions.filter(action=CaseAction.OPEN).first()
+        org = open_action.case.org
+        user = open_action.created_by
+        partner = open_action.case.assignee
+        case = open_action.case
+
+        day = datetime_to_date(open_action.created_on, org)
+        DailyCount.record_item(day, DailyCount.TYPE_CASE_OPENED, org, user)
```
rowanseymour: Should also record at org scope

miltontony:
@rowanseymour this is ready to land - the feedback was addressed in praekelt/casepro#137 which you +1'ed

rowanseymour:
Looks good!

PR #142, opened 2016-08-30
Implement message board in case pro

miltontony:
screen shot 2016-08-29 at 10 20 43

### DIFF
--- a/casepro/msg_board/models.py
+++ b/casepro/msg_board/models.py
@@ -5,6 +5,15 @@ from django_comments.models import Comment
 class MessageBoardComment(Comment):
     pinned_on = models.DateTimeField(null=True, blank=True)
 
+    @classmethod
+    def get_all(cls, org, pinned=False):
+        qs = cls.objects.filter(object_pk=org.pk)
+
+        if pinned:
+            qs = qs.filter(pinned_on__isnull=False)
+
+        return qs
+
     def as_json(self):
         return {
             'comment': self.comment,
--- a/casepro/msg_board/models.py
+++ b/casepro/msg_board/models.py
@@ -1,11 +1,17 @@
-from dash.orgs.models import Org
-
 from django.db import models
-from django.utils.translation import ugettext_lazy as _
+from django_comments.models import Comment
 
 
-class PinnedComment(models.Model):
-    org = models.ForeignKey(Org, verbose_name=_("Organization"), related_name='pinned_comments')
-    comment = models.ForeignKey('django_comments.Comment')
-    pinned_date = models.DateTimeField(auto_now=True)
-    owner = models.ForeignKey('auth.User')
+class MessageBoardComment(Comment):
+    pinned_date = models.DateTimeField(null=True, blank=True)
+    is_pinned = models.BooleanField(default=False)
+
+    def as_json(self):
+        return {
+            'comment': self.comment,
+            'user_name': self.user_name,
+            'user_id': self.user_id,
+            'submit_date': self.submit_date,
+            'pinned_date': self.pinned_date,
+            'comment_id': self.id,
+        }
--- a/casepro/msg_board/models.py
+++ b/casepro/msg_board/models.py
@@ -1,0 +1,11 @@
+from dash.orgs.models import Org
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+
+class PinnedComment(models.Model):
+    org = models.ForeignKey(Org, verbose_name=_("Organization"), related_name='pinned_comments')
+    comment = models.ForeignKey('django_comments.Comment')
+    pinned_date = models.DateTimeField(auto_now=True)
+    owner = models.ForeignKey('auth.User')
--- a/casepro/msg_board/models.py
+++ b/casepro/msg_board/models.py
@@ -3,8 +3,7 @@ from django_comments.models import Comment
 
 
 class MessageBoardComment(Comment):
-    pinned_date = models.DateTimeField(null=True, blank=True)
-    is_pinned = models.BooleanField(default=False)
+    pinned_on = models.DateTimeField(null=True, blank=True)
 
     def as_json(self):
         return {
@@ -12,6 +11,6 @@ class MessageBoardComment(Comment):
             'user_name': self.user_name,
             'user_id': self.user_id,
             'submit_date': self.submit_date,
-            'pinned_date': self.pinned_date,
+            'pinned_on': self.pinned_on,
             'comment_id': self.id,
         }
--- a/casepro/msg_board/models.py
+++ b/casepro/msg_board/models.py
@@ -1,8 +1,8 @@
 from django.db import models
-from django_comments.models import Comment
+from django_comments.models import CommentAbstractModel
 
 
-class MessageBoardComment(Comment):
+class MessageBoardComment(CommentAbstractModel):
     pinned_on = models.DateTimeField(null=True, blank=True)
 
     @classmethod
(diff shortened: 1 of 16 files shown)


miltontony:
@rowanseymour ready for review

casepro/msg_board/models.py
```diff
@@ -0,0 +1,11 @@
+from dash.orgs.models import Org
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+
+class PinnedComment(models.Model):
```
rowanseymour: I think it would be cleaner to add a `is_pinned` field to a custom comment model (https://django-contrib-comments.readthedocs.io/en/latest/custom.html) rather than have two different models

casepro/msg_board/views.py
```diff
@@ -0,0 +1,114 @@
...
+
+class CommentsView(OrgPermsMixin, SmartListView):
+    """
+    JSON endpoint for listing comments
+    """
+    permission = 'orgs.org_inbox'
+    title = 'Message Board'
+
+    def get_queryset(self):
+        return Comment.objects.for_model(Org).filter(object_pk=self.request.org.pk).order_by('-submit_date')
```
rowanseymour: Doesn't `for_model(Org)` already filter by the org ?

casepro/msg_board/models.py
rowanseymour: Other reason for this is not needing two different CRUDLs

pip-freeze.txt
```diff
@@ -57,3 +57,6 @@ wsgiref==0.1.2
 xlrd==0.9.4
 xlutils==1.7.1
 xlwt==1.0.0
+casepropods.dummy
```
rowanseymour: Are these casepropods.* entries supposed to be in this PR ?

rowanseymour:
Some comments on the screenshot:
- Green and blue bubbles are used for input to/from contacts on case timeline. Maybe better to use the same styling as case notes which are input from users.
- Remove spaces between usernames and colon
- Pinning buttons would look better if they we're smaller and on the right after the date.

casepro/msg_board/views.py
miltontony: No, it only filters comments by the content type for `Org` and not `org`

pip-freeze.txt
miltontony: No they're not

miltontony:
@rowanseymour ready for re-review

casepro/msg_board/models.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+from django_comments.models import Comment
+
+
+class MessageBoardComment(Comment):
```
rowanseymour: Should inherit from `CommentAbstractModel` I think

casepro/msg_board/models.py
```diff
@@ -0,0 +1,17 @@
...
+    pinned_date = models.DateTimeField(null=True, blank=True)
```
rowanseymour: Following convention elsewhere for date fields, let's call this `pinned_on`

casepro/msg_board/models.py
```diff
@@ -0,0 +1,17 @@
...
+    is_pinned = models.BooleanField(default=False)
```
rowanseymour: Don't need this field since `pinned_on` can tell us if a comment is pinned

casepro/msg_board/models.py
rowanseymour: Also need to add an org field so messages are filtered by org

rowanseymour: Sorry cancel that - I see that's handled internally by django_comments

casepro/msg_board/views.py
```diff
@@ -0,0 +1,100 @@
...
+class CommentsCRUDL(SmartCRUDL):
+    title = 'Comments'
+    actions = ('list', 'pinned', 'pin', 'unpin')
+    model = MessageBoardComment
+
+    class List(OrgPermsMixin, SmartListView):
+        permission = 'msg_board.messageboardcomment_list'
+        default_order = ('-submit_date',)
+
+        def get_queryset(self):
```
rowanseymour: This should exclude the pinned comments so they don't appear twice (`.filter(pinned_on=None)`)

casepro/msg_board/views.py
```diff
@@ -0,0 +1,100 @@
...
+            return MessageBoardComment.objects.for_model(Org).filter(
```
rowanseymour: For convenience, let's have this be a method on `Org`

```python
def get_message_board_comments(self):
    return MessageBoardComment.objects.for_model(Org).filter(object_pk=self.pk)
```

casepro/msg_board/views.py
```diff
@@ -0,0 +1,100 @@
...
+        fields = ['comment', 'pinned_date']
+        http_method_names = ['post']
+
+        def get_object(self):
+            comment = MessageBoardComment.objects.for_model(Org).filter(
+                object_pk=self.request.org.pk,
+                pk=self.kwargs.get('pk')).first()
+
+            if not comment:
+                raise PermissionDenied
```
rowanseymour: I think more appropriate to use Django's `get_object_or_404`

casepro/msg_board/views.py
miltontony: I disagree. I think this would unnecessarily pollute the `Org` model. Also, since the `Org` model is defined in `dash`, the only way to add this helper method would be to patch it it, which feels unnecessary especially since we only need it in 2 places at the moment.

casepro/msg_board/models.py
miltontony: I was keeping with the convention of `django_comments` that uses `submit_date`

casepro/msg_board/views.py
miltontony: Not quite - the intention for pinned comments is that they remained pinned on the side. The main comments stream should still show the context in which the pinned comments were made. Otherwise for people reading the comments stream after the fact, the messages might not make sense without the pinned comments.

rowanseymour:
Based on screenshots at https://github.com/praekelt/casepro/pull/124, just few UI tweaks needed. 

- There's some spacing now before the username and the user link includes the colon and space after it
- Unpin should move over to the right same as pin
- Can we tone down the pin/unpin buttons a bit?

I checked out the PR and tried tweaking things. Top is how it currently is, bottom is with tweaks above:

screen shot 2016-09-01 at 10 22 45

casepro/msg_board/views.py
rowanseymour: Fair point about having to patch org - we can follow the convention used by other models and have a fetch method that takes the org as param (e.g. `Group.get_all(org, visible=True)`). It's used in 4 places so I think it's worth it, e.g.

`MessageBoardComment.get_all(org, pinned=True)`

casepro/msg_board/views.py
rowanseymour: OK makes sense since they appear off to the side and not just at the top

miltontony:
screen shot 2016-09-01 at 18 15 15

miltontony:
@rowanseymour just pinging you again - I've resolved all the issues raised

casepro/msg_board/models.py
rowanseymour: Base class still needs changed

miltontony: @rowanseymour it doesn't have to. The `Comment` model is a very [light wrapper of the abstract model](https://github.com/django/django-contrib-comments/blob/master/django_comments/models.py#L12-L14). I don't see any reason not to use it.

rowanseymour: It determines the database table used. Since we're creating our own comment subclass, it's better it goes in our own database table in our own app.

casepro/msg_board/models.py
rowanseymour: Comment above also not addressed

PR #145, opened 2016-09-01
Open a case without an inbound message

rudigiesler:
The goal of this is to be able to create a case without having an inbound message from a user.

The use case of this would be if you had a phone line as well as an SMS line, and a phone call comes in, and you need to create a case for the person dialing in. Because they're on a phone line, there's no initial message to be able to create a case from.

The general idea for this is to add a `Create Case` button on the inbox, and allow the user to enter an URN instead of having an initial message. 

This does rely on some of the other PR's, so in the diff there's a few unrelated changes that should go away as PRs are landed and merged in.


rowanseymour:
Replaced by #177

PR #150, opened 2016-09-12
Improve default pod display

KaitCrawford:
This fixes a few problems with the pod styles:
- [x] The default style for pods wasn't being loaded
- [x] When the site is open on a smaller screen the pods appear underneath the timeline and fill the width of the page. However, pod rows then wrap and the row value appears underneath the row name.
- [x] Currently there's no clear separation of rows so rows that wrap can become confusing.


KaitCrawford:
Before:
striping
unstriped_pod
wrapping
unwrapped_pod

KaitCrawford:
After:
striping
striped_pod
wrapping
wrapped_pod

rowanseymour:
This is fine but we should consolidate the styles for non-pod (i.e. the Contact panel) and pod panels - I don't mind if we change the existing non-pod panel style - but right now one uses bold, one uses stripes etc.

PR #154, opened 2016-09-28
Case count tweaks

rowanseymour:
- Adds progress output to migration
- Fixes header case on partner tab of org dashboard


rowanseymour:
@jpaidoussi @miltontony two very minor changes but FYI

PR #160, opened 2016-10-04
Django 1.9 preparation

rowanseymour:
* Updates to latest Django 1.8.x (1.8.15)
* Moves all monkey patch code into models.py files so it doesn't cause errors by accessing models before they are loaded
* Switches to new TEMPLATES setting
* Removes django-reversion which isn't used and version we had isn't compatible with 1.9
* Updates debug-toolbar to a version that works with 1.9


casepro/settings_common.py
```diff
@@ -101,21 +99,6 @@
 # calendars according to the current locale
 USE_L10N = True
 
-# Absolute filesystem path to the directory that will hold user-uploaded files.
-# Example: "/home/media/media.lawrence.com/media/"
-MEDIA_ROOT = ''
```
rowanseymour: These were being redefined below

PR #177, opened 2016-10-17
Open a case without an initial message (v2)

KaitCrawford:
This PR cherry-picks and changes commits from #145 in order to remove it's dependency on #130.
@rowanseymour you reviewed the original PRs.

## Open case without initial message
The goal of this is to be able to create a case without having an inbound message from a user.
The use case of this would be if you had a phone line as well as an SMS line, and a phone call comes in, and you need to create a case for the person dialing in. Because they're on a phone line, there's no initial message to be able to create a case from.
The general idea for this is to add a Create Case button on the inbox, and allow the user to enter an URN instead of having an initial message.


casepro/cases/tests.py
```diff
@@ -860,6 +936,19 @@ def test_timeline(self, mock_fetch_contact_messages):
         mock_fetch_contact_messages.assert_called_once_with(self.unicef, self.ann, d1, case.closed_on)
         mock_fetch_contact_messages.reset_mock()
 
+    def test_timeline_no_initial_message(self):
+        '''If a case has no initial message, the timeline should start from the datetime it was opened.'''
```
rowanseymour: """ for docstrings

casepro/cases/models.py
```diff
@@ -228,37 +228,65 @@ def search(cls, org, user, search):
...
-        with r.lock(CASE_LOCK_KEY % (org.pk, message.contact.uuid)):
-            message.refresh_from_db()
+        contact = message.contact if message else contact
 
-            # if message is already associated with a case, return that
-            if message.case:
-                message.case.is_new = False
-                return message.case
+        with r.lock(CASE_LOCK_KEY % (org.pk, contact.uuid)):
+            if message:
```
rowanseymour: Bit of duplication in the branches of this if statement

static/coffee/services.coffee
```diff
@@ -587,6 +589,58 @@ services.factory('ModalService', ['$rootScope', '$uibModal', ($rootScope, $uibMo
...
+           title, context, initial, maxLength, schemes
+        }),
+        controller: ($scope, $uibModalInstance, PartnerService, UserService) ->
+          $scope.fields = {
+            urn: {scheme: null, path: ""},
+            text: {val: initial, maxLength: maxLength},
+            partner: {val: 0, choices:[]},
+            user: {val: 0, choices: []}
+          }
+
```
rowanseymour: Double spacing

static/coffee/services.coffee
```diff
@@ -587,6 +589,58 @@ services.factory('ModalService', ['$rootScope', '$uibModal', ($rootScope, $uibMo
           $scope.cancel = -> $uibModalInstance.dismiss()
       })
       .result
+
+    create_case: ({
```
rowanseymour: All JS should be camelCase

static/coffee/services.coffee
```diff
@@ -698,3 +752,21 @@ services.factory('MessageBoardService', ['$http', '$httpParamSerializer', '$wind
       return $http.post('/messageboardcomment/unpin/' + comment.id + '/')
 
 ])
+
+#=====================================================================
+# InboxUIService service
+#=====================================================================
+services.factory('InboxUIService', ['CaseService', 'ModalService', 'UtilsService', (CaseService, ModalService, UtilsService) ->
```
rowanseymour: Not convinced this needs to be separate service, rather than in the controller with everything else

static/templates/modals/create_case.html
```diff
@@ -0,0 +1,57 @@
+<div class="modal-header">
```
rowanseymour: Need screenshots

templates/cases/inbox_messages.haml
```diff
@@ -24,6 +24,8 @@
         %form.clearfix
           .btn-group.pull-away
             // TODO figure out why form is double-submitted when this is moved to ng-submit on the form. jQuery conflict?
```
rowanseymour: This comment applies to the original onSearch button line

templates/cases/inbox_messages.haml
```diff
@@ -24,6 +24,8 @@
...
+            %button.btn.btn-default{ type:"button", ng-click:"onCaseWithoutMessage()", ng-if:'folder == "inbox"'}
```
rowanseymour: We need to disable this for now on upartners - so maybe a setting like `SITE_ALLOW_CASE_WITHOUT_MESSAGE`

rowanseymour:
@KaitCrawford have made some comments but also need 1) screenshots and 2) test coverage back at 100%. Thanks!

casepro/contacts/migrations/0025_auto_20161014_1425.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.contrib.postgres.fields
+
+
+class Migration(migrations.Migration):
```
rowanseymour: Can merge this migration with 0023 (or even 0024 as well)

KaitCrawford: addressed in 493ab5e

KaitCrawford:
@rudigiesler I've made some larger changes in the last few commits. Please review the commits after 	6fe340d (the ones made in the last 24hrs essentially) so that someone else on our side has sight of this.

casepro/cases/models.py
KaitCrawford: addressed in c91626a

KaitCrawford:
Create case without message button
createcasebutton
Create case without message modal
createcasemodal

rowanseymour:
UI feedback:
 * "Recipient" doesn't seem like the right word on the modal - maybe "Contact" ?
 * Title "Create case" -> "Open **C**ase" like the existing Open Case modal
 * Likewise modal submit button -> "Open"
 * Don't think the button belongs in the search bar - everything in that bar relates to the messages below. Could put it above, i.e.

screen shot 2016-10-20 at 15 24 01

I notice the padding is different on this modal to other modals, but it appears the other modals aren't setup right - all the modal body is inside the header. We can fix that separately.

KaitCrawford:
@rowanseymour "Recipient" is consistent with the Forward modal:
forward

rowanseymour:
@KaitCrawford sure but in that case it's the "recipient of a message". It doesn't make so much sense to talk about the "recipient of a case"

static/coffee/services.coffee
KaitCrawford: addressed in 88dc12d

KaitCrawford:
inbox
Note that I changed the button above to display on all pages, not just the Inbox page as it previously was.

createcase

rowanseymour:
Looks good - thanks!

PR #178, opened 2016-10-18
Python 3 Support

rowanseymour:


casepro/backend/tests/test_junebug.py
```diff
@@ -285,7 +290,7 @@ def test_outgoing_urn(self):
         msg = self.create_outgoing(self.unicef, self.user1, None, "B", "That's great", bob, urn="tel:+1234")
 
         def request_callback(request):
-            data = json.loads(request.body)
+            data = json_decode(request.body)
```
rowanseymour: Handles decoding the response into unicode

casepro/backend/tests/test_junebug.py
```diff
@@ -1127,11 +1129,11 @@ def request_callback_2(request):
 
         res = identity_store.get_paginated_response(
             ("http://identitystore.org/api/v1/identities/identity-uuid/" "addresses/msisdn"), params={'default': True})
-        self.assertEqual(sorted(res), sorted([
+        self.assertEqual(sorted(res, key=lambda x: x['address']), [
```
rowanseymour: Dicts aren't sortable in Python 3

casepro/cases/tests.py
```diff
@@ -1075,7 +1076,7 @@ def test_create(self):
         self.login(self.admin)
         response = self.url_get('unicef', url)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.context['form'].fields.keys(),
+        self.assertEqual(list(response.context['form'].fields.keys()),
```
rowanseymour: `OrderedDict.keys()` isn't a list in Python 3

casepro/pods/tests/utils.py
```diff
@@ -23,7 +25,7 @@ def perform_action(self, type_, params):
         '''
         Returns a successful action result with a message containing the type and params.
         '''
-        return (True, {'message': 'Type %s Params %r' % (type_, params)})
+        return (True, {'message': 'Type %s Params %s' % (type_, json.dumps(params))})
```
rowanseymour: The `repr` isn't consistent between Python versions so using a JSON dump instead

casepro/profiles/models.py
```diff
@@ -369,5 +369,10 @@ def _user_as_json(user, full=True, org=None):
 User.can_manage = _user_can_manage
 User.can_edit = _user_can_edit
 User.remove_from_org = _user_remove_from_org
-User.__unicode__ = _user_unicode
 User.as_json = _user_as_json
+
+
+if six.PY2:
```
rowanseymour: This class uses the `@python_2_unicode_compatible` decorator so we have to money patch different methods depending on Python version.

casepro/utils/__init__.py
```diff
@@ -59,6 +60,23 @@ def json_encode(data):
     return json.dumps(data, cls=JSONEncoder)
 
 
+def json_decode(data):
+    """
+    Decodes the given JSON as primitives
+    """
+    if isinstance(data, six.binary_type):
+        data = data.decode('utf-8')
```
nicpottier: Can the encoding change based on the header of the request?

See: http://stackoverflow.com/questions/708915/detecting-the-character-encoding-of-an-http-post-request

rowanseymour: There are two places this is used - 1) handling requests made from Angular services and 2) handling responses/requests in the Junebug backend. I'm pretty certain 1) will always use UTF8. For 2) we need to hear from @rudigiesler @jpaidoussi et al.

rudigiesler: Unfortunately we don't send the encoding in the header, but it is possible to add it. The http requests for Junebug are handled here: https://github.com/praekelt/junebug/blob/develop/junebug/workers.py#L245-L249

If I'm remembering correctly, `json.dumps` creates ascii strings with `\uxxxx` strings for unicode characters.

rowanseymour: @rudigiesler you're right about `json.dumps` - I just added some tests to verify that https://github.com/rapidpro/casepro/blob/50977404346206ccbe93a8647761bf5f63d583d4/casepro/utils/tests.py#L92-L92

Do you guys want to try deploying this branch to a test server before it is merged? I was going to likewise to check RapidPro integration.

nicpottier: Shouldn't the "default" be USO unless UTF-8 is in the header?

PR #179, opened 2016-10-19
Remove django-celery as it's not needed

rowanseymour:
And no so well maintained - doesn't have a Python 3 or Django 1.10 compatible release yet.

Need to ensure workers aren't being launched using the management command provided by django-celery, i.e. https://github.com/nyaruka/upartners/pull/2

### DIFF
--- a/casepro/celery.py
+++ b/casepro/celery.py
@@ -10,14 +10,6 @@ os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'casepro.settings')
 
 app = Celery('casepro')
 
-# use django-celery database backend
-app.conf.update(
-    CELERY_RESULT_BACKEND='djcelery.backends.cache:CacheBackend',
-    CELERY_ACCEPT_CONTENT=['json'],
-    CELERY_TASK_SERIALIZER='json',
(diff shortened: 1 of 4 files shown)


nicpottier:
Not sure you use it in casepro but one thing to make sure of is that our transaction management stuff still works. (IE, celery tasks are queued until the db transaction completes)

rowanseymour:
I ripped out django-celery-transactions a while ago as it's no longer maintained. Now we just have `NonAtomicMixin` (from smartmin) for the few views (I think just the 3 export starting views) which actually trigger a celery task based on something they're saving.

nicpottier:
Oh, well nevermind then.

rowanseymour:
@miltontony @rudigiesler @jpaidoussi I don't know what your deploy setup is but you'll want to make sure you're not using the [now removed] django-celery command to launch celery workers, i.e. instead of

`python manage.py celery worker ...`

use

`celery worker -A casepro ...`

rudigiesler:
@rowanseymour Just checked, we're using the newer `celery worker -A casepro...` https://github.com/praekelt/casepro/blob/staging/docker/supervisor.conf#L13

PR #180, opened 2016-10-21
Option to display contact urn instead of contact name

KaitCrawford:
Some sites may prefer to display contact's URN rather than something completely anonymous or the contact's name.


KaitCrawford:
@rudigiesler please give this a review

rudigiesler:
This looks good to me.

casepro/contacts/models.py
```diff
@@ -201,15 +201,24 @@ def lock(cls, org, uuid):
...
+        if getattr(settings, 'SITE_ANON_CONTACTS', False):
             if self.uuid:
                 return self.uuid[:6].upper()
             return ""
-        else:
+        contact_display_format = getattr(settings, 'SITE_CONTACT_DISPLAY', False)
+        if contact_display_format == "name" and self.name:
             return self.name
+        if contact_display_format == "urn" and self.urns:
+            _, path = self.urns[0].split(':', 1)
```
rowanseymour: Best not to use _ as that's usually used as alias for gettext

templates/cases/case_read.haml
```diff
@@ -163,7 +163,10 @@
             .container-fluid
               .row
                 .contact-field-label.col-sm-6
-                  - trans "Name"
+                  - if contact_display_format == 'urn'
+                    - trans "URN"
```
rowanseymour: I don't think most users know what a URN is, so I think here we should list urns with their schemes. As elsewhere, "tel" can become "Phone" etc. If SITE_ANON_CONTACTS is true, then don't list any URNs.

```
Name     Bob                
Phone    +250353252111 
Email    bob@example.com     
```

KaitCrawford: I'm happy to change the display to use the scheme rather than `URN` but changing the display to the example you've given requires a rework of the code since `get_contact_display` is only expected to return a string and really shouldn't be returning a list of urns.

rowanseymour: I think `get_contact_display` should always return a single value, used when a contact needs to be displayed as such. Here I think we'd include URNs in the JSON representation of a contact (provided of course that SITE_ANON_CONTACTS isn't true).

casepro/contacts/tests.py
```diff
@@ -141,6 +141,28 @@ def test_get_display_name(self):
...
+        # if the site uses anon contacts an empty list should be returned
+        with override_settings(SITE_ANON_CONTACTS=True):
+            self.assertEqual(self.ann.get_display_urns(), [])
+
+        # if the contact has no urns an empty list should be returned
+        with override_settings(SITE_CONTACT_DISPLAY="urn"):
+            self.assertEqual(self.ann.get_display_urns(), [])
+
+        # if the contact has urns they should be returned in a list with their display names
+        self.ann.urns = ['tel:+234567890', 'mailto:ann@test.com', 'twitter:@anntest']
```
rowanseymour: I know it's only a test but Twitter URN paths don't include the @

casepro/contacts/models.py
```diff
@@ -430,7 +440,7 @@ def as_json(self, full=True):
         """
         Prepares a contact for JSON serialization
         """
-        result = {'id': self.pk, 'name': self.get_display_name()}
+        result = {'id': self.pk, 'name': self.get_display_name(), 'urns': self.get_display_urns()}
```
rowanseymour: Don't think we should return urns in the compact view of a contact

casepro/contacts/tests.py
```diff
@@ -175,16 +198,33 @@ def test_as_json(self):
...
-        # if site uses anon contacts then name is obscured
+        # If the contact has urns and the contact display uses urns then they should be returned and the name should be
+        # the first urn
+        self.ann.urns = ['tel:+234567890', 'mailto:ann@test.com']
+        self.ann.save()
+        with override_settings(SITE_CONTACT_DISPLAY='urn'):
+            self.assertEqual(self.ann.as_json(full=True), {
+                'id': self.ann.pk,
+                'name': "+234567890",
+                'urns': [{'scheme': "Phone", 'path': "+234567890"}, {'scheme': "Email", 'path': "ann@test.com"}],
```
rowanseymour: I think better to just return full URN strings here (i.e. "tel:+234567890"). We can split them in the frontend code and just use a hardcoded map to look up scheme names from scheme prefixes.

Otherwise if we made a contact editing form we'd need to translate these scheme names back into scheme prefixes. Also when a user submits a Forward to URN, we treat the URNs as full strings so good if this is consistent with that.

casepro/contacts/models.py
KaitCrawford: The compact view is the representation that's currently available on the case page which is one of the places we want the urns to be visible. I'll change the case_read view to retrieve the urns instead.

rowanseymour: The case read page fetches the full contact in a separate request, no? https://github.com/rapidpro/casepro/blob/8db6ef3db6736697d46d47b62ee2e2a517ea034d/static/coffee/controllers.coffee#L587-L587

KaitCrawford: Ah I see. I've been using the case object to get all of the contact info. Thanks

KaitCrawford:
@rowanseymour This is ready for re-review

templates/cases/case_read.haml
```diff
@@ -161,11 +161,16 @@
...
+                .contact-field-label.col-sm-6
+                  [[ urn.scheme ]]
+                .contact-field-value.col-sm-6
+                  [[ urn.path ]]
+              .row{ ng-hide:"contact.urns.length > 0" }
                 .contact-field-label.col-sm-6
                   - trans "Name"
                 .contact-field-value.col-sm-6
-                  [[ caseObj.contact.name ]]
+                  [[ contact.name ]]
```
rowanseymour: For your usecase do you need to intentionally hide name even if it's set? If not I think it should be displayed even when a contact has urns - and I think be the top item.

rowanseymour: You mean name has be hidden for anonymity reasons, or it's just not used?

KaitCrawford: It has to be hidden for anonymity reasons.

rowanseymour: Ok so taking step back here, there are two things we want deployments to be able to customize:

1. What contact fields are allowed to be displayed
2. How a contact is displayed as single value

For our different requirements that means:

* For UPartners we want 1) to hide both URNS and name, and 2) to use an anon identifier based on UUID
* For Momconnect you need 1) to hide names and 2) the use a URN
* Another deployment might not have of our requirements and would 1) likely want to display all fields and 2) render contacts with their names - so this I think should be the default for CasePro.

To do that all of that I think we need to replace `SITE_ANON_CONTACTS` with something that handle 1), like `SITE_HIDE_CONTACT_FIELDS = ('name', ...)`. That setting determines what goes into the full JSON representation of a contact, and what we display on the contact read page. The name in the full representation should be the actual name unless `SITE_HIDE_CONTACT_FIELDS` contains name.

To make things simpler we can have an new field in the contact JSON representation called `display` whose content is determined by `SITE_CONTACT_DISPLAY`. For example the compact representation of a contact might look like:

```json
{"id": 1234, "display": "11EDA2"}  // UPartners
{"id": 1234, "display": "+25032432544"}  // Momconnect
{"id": 1234, "display": "Bob Jones"}  // Default
```

And the full representation...
```json
{"id": 1234, "display": "11EDA2", "name": null, "urns": [], ...}  // UPartners
{"id": 1234, "display": "+25032432544", "name": null, "urns": ["tel:+25032432544"], ...}  // Momconnect
{"id": 1234, "display": "Bob Jones", "name": "Bob Jones", "urns": ["tel:+25032432544"], ...}  // Default
```

I think the separate `display` field keeps things simpler and the introduction of `SITE_HIDE_CONTACT_FIELDS` will be helpful when we get to the point of wanting to edit contacts etc. Thoughts?

KaitCrawford: @rowanseymour I've implemented your suggestion in the latest commits.

rowanseymour:
Looks good! Thanks for your patience on this one.

PR #181, opened 2016-10-25
Contact URN tweaks

rowanseymour:
* Moves URN functionality into URN class
* Add support for normalizing URNs
* Fixes open case modal so required error not shown on Summary field unless form has been submitted

@KaitCrawford currently on the 'Open Case' modal, the user has to type a phone number in exactly E164 format, and sees an error message as they type. This PR doesn't address that but does add normalization on the server side, so it might be possible to relax that restriction and let users type things like "260 96-123-1234"

### DIFF
--- a/casepro/contacts/tests.py
+++ b/casepro/contacts/tests.py
@@ -13,7 +13,46 @@ from .tasks import pull_contacts
 
 
 class URNTest(BaseCasesTest):
-    def test_validate_urn(self):
+    def test_from_parts(self):
+        self.assertEqual(URN.from_parts("tel", "12345"), "tel:12345")
+        self.assertEqual(URN.from_parts("tel", "+12345"), "tel:+12345")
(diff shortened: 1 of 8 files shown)


casepro/contacts/tests.py
```diff
@@ -122,11 +175,13 @@ def test_as_json(self):
             self.assertEqual(self.ann.as_json(full=False), {'id': self.ann.pk, 'name': "7B7DD8"})
 
     @patch('casepro.test.TestBackend.push_contact')
-    def test_get_or_create_from_urn_no_match(self, mock_push_contact):
+    def test_get_or_create_from_urn(self, mock_push_contact):
```
rowanseymour: We tend to use one test method per real method because all of our unit tests are integration tests and each new method requires a db transaction to be created and rolled back.

PR #184, opened 2016-11-03
Ensure inbox header buttons are displayed as group

rowanseymour:
Before

screen shot 2016-11-03 at 11 10 00

After

screen shot 2016-11-03 at 11 08 55


rowanseymour:
@KaitCrawford FYI

PR #186, opened 2016-11-07
fix urn display css

miltontony:


KaitCrawford:
Urns displayed on the inbox page are displaying poorly due to a lack of space.
This adjusts the css to improve the display.
Before changes:
urns display

After changes:
new_display

miltontony:
@rowanseymour small review please

rowanseymour:
@miltontony I think it looks better when labels are aligned. 

For now, you guys are always going to display a full telephone number, and we'll always display the six character anon id. So I think we can have switch the width of that column depending on SITE_CONTACT_DISPLAY. That way it will be unchanged for us, but you guys will get more space for phone numbers

miltontony:
Closing this and replacing it with #209

PR #189, opened 2016-11-11
Update to latest dask and add lock timeouts on long running org tasks

rowanseymour:


casepro/contacts/tasks.py
```diff
@@ -6,7 +6,7 @@
 logger = get_task_logger(__name__)
 
 
-@org_task('contact-pull')
+@org_task('contact-pull', lock_timeout=12 * 60 * 60)
```
rowanseymour: No sync task should ever take 12 hours

PR #190, opened 2016-11-15
10second refresh on inbox

projectmushroom:
Added a 10second refresh on inbox messages.

Multiple operators are logged in at the same time and currently replying to the same messages, they need as "real time" refresh as possible.

### DIFF
--- a/pip-freeze.txt
+++ b/pip-freeze.txt
@@ -5,7 +5,7 @@ Pygments==2.0.2
 amqp==1.4.7
 anyjson==0.3.3
 billiard==3.3.0.22
-boto==2.38.0
+boto==2.43.0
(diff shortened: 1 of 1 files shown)


static/coffee/controllers.coffee
```diff
@@ -270,7 +270,18 @@ controllers.controller('MessagesController', ['$scope', '$timeout', '$uibModal',
     )
 
   $scope.fetchOldItems = (search, startTime, page) ->
-    return MessageService.fetchOld(search, startTime, page)
+    $scope.autoRefresh = ->
```
rowanseymour: This doesn't look like it can work with multiple pages - different pages will overwrite all of $scope.items

PR #193, opened 2016-11-16
Incorrect value shown for total cases in stags

miltontony:
image


miltontony:
@rowanseymour small review please

PR #194, opened 2016-11-16
additional test for cases that were created manually

miltontony:


casepro/backend/junebug.py
```diff
@@ -34,12 +34,19 @@ def __init__(self, base_url, auth_token):
         self.session.headers.update({'Content-Type': "application/json"})
 
     def build_outgoing_message_json(self, outgoing, to_addr):
+        if not outgoing.reply_to:
+            reply_to = ''
+            label = ''
+        else:
+            reply_to = reply_to = outgoing.reply_to.text
```
rudigiesler: Double `reply_to =`?

casepro/backend/junebug.py
```diff
@@ -34,12 +34,19 @@ def __init__(self, base_url, auth_token):
...
+            label = ','.join([str(l) for l in outgoing.reply_to.labels.all()])
```
rudigiesler: Since it's JSON, we could probably just send this as a list instead of a string? But that would also mean changes to the receiving side, so up to you.

miltontony: Happy to leave this as is.

PR #195, opened 2016-11-17
send created_by__id to junebug hub

miltontony:
We need to pass the Outgoing message creator to Junebug Hub


miltontony:
@rowanseymour review please.

Issue #199, opened 2016-11-18
Reply modal isn't affected by whether user must use FAQs


rowanseymour:

rowanseymour:
Fixed in #200

PR #200, opened 2016-11-18
FAQ fixes and UI tweaks

rowanseymour:
* Makes styling more consistent with rest of app
* Fixes bug where reply modal isn't affected by whether user must use FAQs #199 
* Improves instructions on reply modal based on what user can do

Before:
screen shot 2016-11-18 at 11 00 10

After:
screen shot 2016-11-18 at 11 01 09

Before:
screen shot 2016-11-18 at 11 00 21

After:
screen shot 2016-11-18 at 11 00 52

Before:
screen shot 2016-11-18 at 10 59 43

After (when user isn't limited to FAQs and there are some):
screen shot 2016-11-18 at 11 12 26

After (when user isn't limited to FAQs and there aren't any):
screen shot 2016-11-18 at 10 56 07

After (when user is limited to FAQs and there are some):
screen shot 2016-11-18 at 11 13 01

After (when user is limited to FAQs and there aren't any):
screen shot 2016-11-18 at 10 55 34

### DIFF
--- a/casepro/profiles/context_processors.py
+++ b/casepro/profiles/context_processors.py
@@ -8,11 +8,14 @@ def user(request):
     if request.user.is_anonymous() or not request.org:
         is_admin = False
         partner = None
+        is_faq_only = True
     else:
         is_admin = request.user.can_administer(request.org)
         partner = request.user.get_partner(request.org)
+        is_faq_only = request.user.must_use_faq()
 
     return {
         'user_is_admin': is_admin,
-        'user_partner': partner
+        'user_partner': partner,
+        'user_is_faq_only': is_faq_only
     }
--- a/templates/msgs/faq_list.haml
+++ b/templates/msgs/faq_list.haml
@@ -9,17 +9,17 @@
     .clearfix
       .page-header-buttons
         .btn-group
-          %a.btn.btn-default{ href:"{% url 'msgs.faq_import' %}" }
-            %span.glyphicon.glyphicon-duplicate
-            - trans "Batch Upload"
-          %a.btn.btn-default{ ng-click:"onNewFaq(faq)", tooltip:"New FAQ" }
+          %a.btn.btn-default{ ng-click:"onNewFaq(faq)", tooltip:"Add new FAQ" }
             %span.glyphicon.glyphicon-plus
             - trans "Add"
+          %a.btn.btn-default{ href:"{% url 'msgs.faq_import' %}" }
+            %span.glyphicon.glyphicon-import
+            - trans "Import"
       .page-header
         %h2
           - trans "FAQs"
       
-      %table.table.table-striped.list-table.table-bordered
+      %table.table.table-striped.list-table
         %thead
           %tr
             %th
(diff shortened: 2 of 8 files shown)


casepro/profiles/context_processors.py
```diff
@@ -8,11 +8,14 @@ def user(request):
...
     else:
         is_admin = request.user.can_administer(request.org)
         partner = request.user.get_partner(request.org)
+        is_faq_only = request.user.must_use_faq()
 
     return {
         'user_is_admin': is_admin,
-        'user_partner': partner
+        'user_partner': partner,
+        'user_is_faq_only': is_faq_only
```
rowanseymour: I get now why this was originally a context processor because otherwise there's no easy way to pass it to the reply modal template

templates/msgs/faq_list.haml
```diff
@@ -32,18 +32,17 @@
...
-                %a{ ng-href:"/faq/read/[[ faq.id ]]/" }
-                  [[ faq.question ]]
-              %td.col-xs-6
-                [[ faq.answer ]]
-              %td
-                [[ faq.language.name ]]
-              %td.text-center
-                %span.badge
-                  [[ faq.count ]]
+          %tr{ ng-repeat:"faq in faqs | filter:filterParents" }
```
rowanseymour: This seems a bit inefficient but I've left it as is for now. Would be better to only fetch non-translation FAQ instances on the server

templates/partials/modal_reply.haml
```diff
@@ -3,53 +3,55 @@
 .modal-header.no-border
   %h3.modal-title
     - trans "Reply"
-  .modal-body.no-bottom-padding{ ng-init:'init()' }
```
rowanseymour: Note the .modal-body here is nested inside the .modal-header

templates/partials/modal_reply.haml
```diff
@@ -3,53 +3,55 @@
...
+            %button.btn.btn-default.dropdown-toggle{ type:"button", uib-dropdown-toggle:"true" }
+              [[lang]]
+              %span.caret
+            %ul.dropdown-menu
+              %li{ ng-repeat:'language in languages' }
+                %a{ ng-click:"filterByLanguage(language)" }
+                  [[language.name]]
 
+      .{ style:"overflow-y:scroll; max-height:12em;" }
+        .stackitem.clearfix.hoverable{ ng-repeat:"reply in replies | filter:searchFaqs", ng-click:"setResponse(reply.answer)" }
```
rowanseymour: Makes it so user can select the whole item rather than just the item text

rowanseymour:
Please review @miltontony @jpaidoussi @projectmushroom

templates/partials/modal_reply.haml
projectmushroom: I've noticed this on all the other modals as well, it works/looks much better when not nested

projectmushroom:
👍 from my side, looks good, the more condensed modal works much better

templates/partials/modal_reply.haml
rowanseymour: Yeah it's one of those mistakes that's been copy+pasted everywhere

casepro/profiles/context_processors.py
miltontony: 👌

Issue #202, opened 2016-11-18
Problems in response timing calculations


rowanseymour:
* Only counts cases that have been re-assigned to a partner. Should count cases that have been re-assigned, or assigned at opening time to the partner, but exclude cases they assigned to themselves. See https://github.com/rapidpro/casepro/blob/e225631591041ea8cea7dd16a99d16cb1a3e84fe/casepro/statistics/signals.py#L57-L57
* Migration to calculate timings on existing data doesn't quite match signal logic - it wasn't updated to reflect this change to the signal logic https://github.com/rapidpro/casepro/pull/138/commits/e225631591041ea8cea7dd16a99d16cb1a3e84fe

miltontony:
@rowanseymour this can be closed.

PR #203, opened 2016-11-18
Change case referral counts in migration

miltontony:


miltontony:
@rowanseymour review please
This addresses the second point in #202

rowanseymour:
@miltontony it does but I think the logic isn't quite right for both the signal and the migration, so better to fix both here

PR #207, opened 2016-11-28
Issue 196 cant edit faq translations.

erikh360:
The labels should be a empty list instead of None.


casepro/msgs/forms.py
```diff
@@ -121,6 +121,9 @@ def clean_labels(self):
         if parent is None and labels is None:
             raise forms.ValidationError(_("Labels are required if no Parent is selected"))
 
+        if labels is None:
```
rowanseymour: This is kinda odd because `labels` is assigned `None` just above. I think it would be cleaner to use `derive_fields` on the view to remove labels as a field altogether when editing a translation.

erikh360:
@rowanseymour please have a look again.

PR #208, opened 2016-11-28
Issue 197 faq translations dont appear with labels on the reply modal

erikh360:
Added parent labels if FAQ is a translations.

### DIFF
--- a/casepro/msgs/models.py
+++ b/casepro/msgs/models.py
@@ -234,6 +234,7 @@ class FAQ(models.Model):
             queryset = queryset.filter(Q(question__icontains=text) | Q(answer__icontains=text))
 
         queryset = queryset.prefetch_related('labels')
+        queryset = queryset.prefetch_related('parent__labels')
 
         return queryset.order_by('question')
 
@@ -269,9 +270,7 @@ class FAQ(models.Model):
                 result['labels'] = [l.as_json() for l in self.labels.all()]
             else:
                 parent_json = self.parent.id
-
-                labels = FAQ.objects.prefetch_related('labels').get(pk=parent_json).labels.all()
-                result['labels'] = [l.as_json() for l in labels]
+                result['labels'] = [l.as_json() for l in self.parent.labels.all()]
 
             result['answer'] = self.answer
             result['language'] = self.get_language()
--- a/casepro/msgs/models.py
+++ b/casepro/msgs/models.py
@@ -269,7 +269,9 @@ class FAQ(models.Model):
                 result['labels'] = [l.as_json() for l in self.labels.all()]
             else:
                 parent_json = self.parent.id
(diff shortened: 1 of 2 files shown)


casepro/msgs/models.py
```diff
@@ -266,13 +266,14 @@ def as_json(self, full=True):
         if full:
             if not self.parent:
                 parent_json = None
+                result['labels'] = [l.as_json() for l in self.labels.all()]
             else:
                 parent_json = self.parent.id
+                result['labels'] = [l.as_json() for l in self.parent.labels.all()]
```
rowanseymour: Can we prefetch `parent__labels` to minimize db hits?

erikh360:
@rowanseymour please have a look again.

casepro/msgs/models.py
```diff
@@ -266,13 +266,16 @@ def as_json(self, full=True):
         if full:
             if not self.parent:
                 parent_json = None
+                result['labels'] = [l.as_json() for l in self.labels.all()]
             else:
                 parent_json = self.parent.id
 
+                labels = FAQ.objects.prefetch_related('labels').get(pk=parent_json).labels.all()
```
rowanseymour: This doesn't help because you're still separately fetching parent labels for each FAQ. The idea is to prefetch somewhere like `FAQ.get_all()` so that only one db hit is used to fetch parent labels for all the FAQs being fetched.

Best way to test stuff like this is fetch and serialize a bunch of FAQs together and use `with self.assertNumQueries(N):` to check that you're not making db queries for each FAQ being returned.

erikh360:
@rowanseymour please review again, thanks.

casepro/msgs/models.py
```diff
@@ -234,6 +234,7 @@ def search(cls, org, user, search):
             queryset = queryset.filter(Q(question__icontains=text) | Q(answer__icontains=text))
 
         queryset = queryset.prefetch_related('labels')
+        queryset = queryset.prefetch_related('parent__labels')
```
rowanseymour: queryset = queryset.prefetch_related('labels', 'parent__labels')

casepro/msgs/tests.py
```diff
@@ -571,7 +571,9 @@ def test_search(self):
         # log in as a non-administrator
         self.login(self.user1)
 
-        response = self.url_get('unicef', url, {})
+        # check that appropriate number of queries are executed
+        with self.assertNumQueries(28):
```
rowanseymour: Hmm seems high. Can you check adding another FAQ (temporarily) and verify that this number doesn't increase? If it does then something is not quite right.

casepro/msgs/models.py
erikh360: Fixed this, ready for review @rowanseymour

PR #211, opened 2016-12-01
Character limit issue on cases and outgoing messages.

erikh360:
This moves the character limits into the settings_common.py.

Fields size limits are from db and will be handled by frontend.


erikh360:
@rowanseymour please review.

casepro/cases/models.py
```diff
@@ -153,7 +153,7 @@ class Case(models.Model):
 
     initial_message = models.OneToOneField(Message, null=True, related_name='initial_case')
 
-    summary = models.CharField(verbose_name=_("Summary"), max_length=255)
+    summary = models.TextField(verbose_name=_("Summary"))
```
rowanseymour: The point of the summary field is to provide a *short* description of the case which can be used when listing cases. If users are trying to do something else with it, then maybe we're missing something, but removing this limit is not the solution.

casepro/msgs/models.py
```diff
@@ -303,7 +303,7 @@ class Message(models.Model):
 
     type = models.CharField(max_length=1)
 
-    text = models.TextField(max_length=640, verbose_name=_("Text"))
+    text = models.TextField(verbose_name=_("Text"))
```
rowanseymour: As I said in dev list discussion - we can increase this but we should have *a* hard limit. External systems need to know what the absolute longest message they might receive this. Do you guys have a need to be sending anything longer than say 800 (5 concatenated 7bit SMSs) ?

casepro/msgs/models.py
```diff
@@ -608,7 +608,7 @@ class Outgoing(models.Model):
 
     activity = models.CharField(max_length=1, choices=ACTIVITY_CHOICES)
 
-    text = models.TextField(max_length=640)
+    text = models.TextField()
```
rowanseymour: As above.

casepro/settings_common.py
```diff
@@ -56,6 +56,8 @@
 SITE_HIDE_CONTACT_FIELDS = []  # Listed fields should not be displayed
 SITE_CONTACT_DISPLAY = "name"  # Overrules SITE_HIDE_CONTACT_FIELDS Options: 'name', 'uuid' or 'urns'
 SITE_ALLOW_CASE_WITHOUT_MESSAGE = True
+MAX_MESSAGE_CHARS = 140
```
rowanseymour: Thus far we've prefixed CasePro specific settings with SITE_ so let's continue with that.

erikh360:
@rowanseymour I have made the changes requested, please review.

PR #212, opened 2016-12-06
Archive: Show Labels

projectmushroom:
*Requirements*
- Helpdesk staff would also like to see the label ‘population’ of the archive, similar to Inbox. 
- Suggestion: Show labels underneath Archive as well in order to show how many messages have been responded to within each label. 
- They would also like to click on a label under Archive in order to view all ‘Compliments’ that have been archived, for example.
- the labels under Inbox and Archive should be collapsable

screen shot 2016-12-06 at 10 49 30 am
screen shot 2016-12-06 at 10 49 17 am
screen shot 2016-12-06 at 10 50 30 am

### DIFF
--- a/casepro/msgs/models.py
+++ b/casepro/msgs/models.py
@@ -356,7 +356,7 @@ class Message(models.Model):
         all_label_access = user.can_administer(org)
 
         if all_label_access:
-            if folder == MessageFolder.inbox:
+            if folder == MessageFolder.inbox or (folder == MessageFolder.archived and label_id):
                 queryset = queryset.filter(has_labels=True)
                 if label_id:
                     label = Label.get_all(org, user).filter(pk=label_id).first()
--- a/casepro/backend/junebug.py
+++ b/casepro/backend/junebug.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 from datetime import datetime
+import dateutil.parser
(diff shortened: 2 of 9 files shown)


projectmushroom:
@rowanseymour Please review

casepro/msgs/models.py
```diff
@@ -356,7 +356,7 @@ def search(cls, org, user, search):
         all_label_access = user.can_administer(org)
 
         if all_label_access:
-            if folder == MessageFolder.inbox:
+            if folder == MessageFolder.inbox or (folder == MessageFolder.archived and label_id):
```
rowanseymour: Not getting why `and label_id` is here. Isn't this the path we want when an admin is viewing all archived messages?

rowanseymour:
I forgot we already track archived counts for labels - certainly makes this a lot simpler. Looks good - just one question on the code which might be me forgetting how things work.

casepro/msgs/models.py
projectmushroom: We're using label_id because it only need to filter with labels when you select a label in the Archived submenu, when you click on Archived it shouldn't use the has_labels filter

rowanseymour: Ah right because Archived includes unlabelled messages for admins

PR #220, opened 2017-01-19
inbox refresh and message locking

projectmushroom:
- Lock message whilst it's being worked on
- 10 second message refresh interval 
- 5 minute message locking
- Message lock triggered when opening the following modals: Reply, Bulk Reply, Forward, Open Case
- Message in a folder will either be locked or added/removed from of current folder, depending on the action
- When User-A has an open modal (Reply, Bulk Reply, Forward, Open Case), User-B's messages will be locked after the 10 second refresh interval
- When User-A dismisses/completes the modal, User-B's message will be unlocked or added/removed after the refresh interval
- The secondary alert message will only be visible when a user tries to action a currently busy message before the refresh interval runs.
- Message lock also triggered when entering a open case, and unlocked when exiting

screen shot 2017-01-06 at 10 48 51 am


casepro/msgs/models.py
```diff
@@ -384,14 +391,19 @@ def search(cls, org, user, search):
...
             queryset = queryset.filter(is_flagged=True)
 
-        # archived messages can be implicitly or explicitly included depending on folder
-        if folder == MessageFolder.archived:
-            queryset = queryset.filter(is_archived=True)
-        elif folder == MessageFolder.flagged:
-            if not include_archived:
-                queryset = queryset.filter(is_archived=False)
+        # if this is a refresh we want everything with new actions
+        if last_refresh:
```
projectmushroom: We've tested it on flagged view, and it worked as expected.

projectmushroom:
@rowanseymour we've also done testing on a large database, all good

casepro/msgs/models.py
rowanseymour: If a user has the flagged view open, and another user unflags a message, it won't be removed in the first user's view - because you're only fetching flagged messages you can't see an updated message which has been unflagged.

casepro/cases/views.py
```diff
@@ -69,7 +69,8 @@ def get_context_data(self, **kwargs):
             # angular app requires context data in JSON format
             context['context_data_json'] = json_encode({
                 'all_labels': [l.as_json() for l in labels],
-                'fields': [f.as_json() for f in fields]
+                'fields': [f.as_json() for f in fields],
+                'message_id': case.initial_message.backend_id
```
rowanseymour: Not getting why is needed here - can't this just be added to the case JSON representation?

casepro/msgs/views.py
```diff
@@ -216,11 +233,62 @@ def get_context_data(self, **kwargs):
...
+                    for message in messages:
+                        message.locked_on = timezone.now()
+                        message.locked_by = user
+                        message.save()
+
+            elif action == 'unlock':
+                for message in messages:
+                    message.locked_on = None
+                    message.locked_by = None
+                    message.save()
```
rowanseymour: Any time you call `.save()` to update a few fields, you should specify those fields in the `update_fields` arg. Minimises chances of writing stale data

projectmushroom:
@rowanseymour we've made the changes, also fixed the flagged view, it now updates unflagged messages correctly

static/coffee/controllers.coffee
```diff
@@ -604,6 +714,11 @@ controllers.controller('CaseController', ['$scope', '$window', '$timeout', 'Case
       )
     )
 
+  # unlock message on exit
+  $window.onbeforeunload = ->
+    MessageService.checkLock($scope.caseObj.msg_backend_id, true)
```
rowanseymour: I'm not convinced this is useful - either it should be locking all messages that belong to a case or none of them - and I think it's ok to not lock them whilst a case is being viewed since the user can see that those messages are associated with a case anyway - and they've been archived.

casepro/cases/models.py
```diff
@@ -458,7 +458,8 @@ def as_json(self, full=True):
                 'labels': [l.as_json(full=False) for l in self.labels.all()],
                 'summary': self.summary,
                 'opened_on': self.opened_on,
-                'is_closed': self.is_closed
+                'is_closed': self.is_closed,
+                'msg_backend_id': self.initial_message.backend_id if self.initial_message else None,
```
rowanseymour: This should be called `initial_message` since cases have many messages and also the fact that we use backend ids is a hangover from the past that we should fix so I wouldn't put that in the name.

It should also be prefetched to avoid repeat db hits when listing cases.

However see subsequent comments about whether this field is needed.

static/coffee/services.coffee
```diff
@@ -154,6 +157,19 @@ services.factory('MessageService', ['$rootScope', '$http', '$httpParamSerializer
...
 
       return $http.post('/message/action/' + action + '/', params)
+
+    #----------------------------------------------------------------------------
+    # Check if message is locked, set locked state and return locked message ids
+    #----------------------------------------------------------------------------
+    checkLock: (items, unlock) ->
+      action = if unlock then 'unlock' else 'lock'
+      params = {
+        messages: if angular.isArray(items) then (i.id for i in items) else [items]
```
rowanseymour: It's somewhat confusing that this method takes an array of objects or a single id. Better if it always takes an array of msg objects, so I would change the JSON representation of a case to include initial_message like 
```
initial_message: {id: 123, text: "Hello"}
```

However as I've said above, I'm not sure we need to include initial message with a case but I'm not sure we need to lock that message whilst a case is being viewed.

templates/frame.haml
```diff
@@ -203,6 +203,9 @@
 
       function displayAlert(type, text) {
         $('.content').prepend('<div class="alert alert-' + type + '"><a class="close" data-dismiss="alert">×</a>' + text + '</div>');
+        $(".alert").delay(4000).fadeOut(500, function() {
+          $(this).alert('close');
+        });
```
rowanseymour: This hides all alerts after 4 seconds, right? Is this something that users have requested?

projectmushroom:
@rowanseymour As you've mentioned, the message lock is not really necessary since the user can see that those messages are associated with a case anyway. We've removed the message lock/unlock from the existing case view also the auto fade of notices.

casepro/msgs/models.py
```diff
@@ -417,6 +429,15 @@ def get_history(self):
         """
         return self.actions.select_related('created_by', 'label').order_by('-pk')
 
+    def get_lock(self, user_id):
```
rowanseymour: Sorry for not noticing this in last review but why is this an id rather than a user object?

erikh360: Hi @rowanseymour, we only use the ID in the function, and in one case we get the ID from context, https://github.com/praekelt/casepro/blob/6beca26817e493f388b6310be319aa97e2808d8b/casepro/msgs/views.py#L241, so we don't have the object.

rowanseymour: I'm pretty sure the user object is always in the context thanks to the `auth` context processor (i.e. `context['user']`). And even if it wasn't, you could just put the user object rather than the id in at https://github.com/rapidpro/casepro/pull/220/files#diff-23f5fc0c123c019ee596bef56b7f1e89R211.

At the model level, it's better to be passing around objects rather than database ids where possible.

erikh360: Ok thanks, then it shouldn't be a problem.

casepro/msgs/models.py
```diff
@@ -417,6 +429,15 @@ def get_history(self):
...
+        if self.locked_on and self.locked_by:
+            if self.locked_on > (now() - timedelta(seconds=MESSAGE_LOCK_SECONDS)):
+                if self.locked_by.id != user_id:
```
rowanseymour: Also use `locked_by_id` or ensure `locked_by` is prefetched to avoid db hits when calling this on a list of messages. Maybe include a `self.assertNumQueries` check in the tests to make sure we're not missing any prefetches.

projectmushroom:
@rowanseymour we've made the necessary changes

casepro/msgs/tests.py
```diff
@@ -1211,6 +1211,78 @@ def test_search(self):
...
+        self.assertEqual(len(response.json['results']), 2)
+        self.assertEqual(response.json['results'][0]['id'], 210)
+        self.assertEqual(response.json['results'][1]['id'], 105)
+        self.assertEqual(response.json['results'][1]['flagged'], True)
+
+    def test_get_lock(self):
+        msg = self.create_message(self.unicef, 101, self.ann, "Normal", [self.aids, self.pregnancy])
+
+        # The message is not locked
+        with self.assertNumQueries(0):
```
rowanseymour: This isn't really checking anything since self.user1 is obviously fetched. What I meant was checking that when we fetch multiple messages over the search endpoint, we don't have to db hits for each message because we haven't prefetched locked_by. You can remove these and I'll get this merged.

projectmushroom:
@rowanseymour thanks Rowan!

PR #224, opened 2017-05-30
Delete personal info from optout hook

erikh360:


erikh360:
@rowanseymour Please review, thanks!

PR #239, opened 2017-11-20
Inbox auto-refresh fix

rowanseymour:


casepro/msgs/migrations/0059_new_modified_msgs_index.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+SQL = """
+CREATE INDEX CONCURRENTLY msgs_message_org_modified_on
+ON msgs_message(org_id, has_labels, modified_on, created_on DESC)
+WHERE is_active = TRUE AND is_handled = TRUE;
+"""
```
rowanseymour: Adding a better index here for fetching modified messages

casepro/msgs/tasks.py
```diff
@@ -79,7 +79,7 @@ def handle_messages(org):
         rule_processor.apply_actions()
 
         # mark all of these messages as handled
-        Message.objects.filter(pk__in=[m.pk for m in unhandled]).update(is_handled=True)
+        Message.objects.filter(pk__in=[m.pk for m in unhandled]).update(is_handled=True, modified_on=timezone.now())
```
rowanseymour: Handling now sets modified_on so we can use that to fetch new/modified messages

static/coffee/controllers.coffee
```diff
@@ -241,8 +241,9 @@ controllers.controller('BaseItemsController', ['$scope', '$timeout', 'UtilsServi
           item.selected = $scope.items[scopeItems[item.id]].selected
           $scope.items[scopeItems[item.id]] = item
         else
-          # new item so we add it to the top
-          $scope.items.unshift(item)
+          $scope.items.push(item)
+
+      $scope.items = $filter('orderBy')($scope.items, 'time', true)
```
rowanseymour: We might pick up an old message which has reappeared in this view

casepro/msgs/migrations/0059_new_modified_msgs_index.py
ericnewcomer: Is there a similar index that can now be dropped?

rowanseymour: There's a Django managed index that gets dropped after this is created by setting index_together to none

PR #242, opened 2018-01-23
Custom through model for Message/Label m2m

rowanseymour:
Doesn't add any new fields - just gets us to a place where we could


casepro/msgs/migrations/0012_message_labels.py
```diff
@@ -11,9 +12,28 @@ class Migration(migrations.Migration):
...
 
     operations = [
+        migrations.CreateModel(
+            name='Labelling',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('label', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='msgs.Label')),
+            ],
+            options={
+                'db_table': 'msgs_message_labels',
```
rowanseymour: This is the "rewrite migration history to make it look like it always was a custom model" approach

PR #243, opened 2018-01-30
Search optimization part 1

rowanseymour:
Adds `Labelling.message_created_on`


nicpottier:
So!? How much faster is it?

rowanseymour:
Hard to say because I can't replicate the performance problems locally (maybe I need to run a copy of ureport in the same db 😉).. it does tho make for a much nicer query plan without the in-memory sort.

But also I can't get it to use that plan unless I change `seq_page_cost` to make psql think that fetching pages is slower than it really is. So still experimenting...

pip-freeze.txt
```diff
@@ -27,7 +27,7 @@ django-redis==4.8.0
 django-rosetta==0.7.13
 django-storages==1.5.2
 django-timezone-field==2.0  # via rapidpro-dash
-django==1.11.7
+django==1.11.9
```
rowanseymour: to get change by @dodobas that makes Django user server-side cursors with values_list (needed for next part)

PR #249, opened 2018-03-08
Update rapidpro-dash 

norkans7:
-Migrations for org config
-Use OrgBackend object


norkans7:
This will be ready once https://github.com/rapidpro/dash/pull/105 is merged and here we use the latest release

PR #356, opened 2021-08-02
🚀 Update to Django 3.2 LTS

rowanseymour:


casepro/backend/tests/test_rapidpro.py
```diff
@@ -418,7 +418,7 @@ def test_pull_contacts(self, mock_get_contacts):
             ),
         ]
 
-        with self.assertNumQueries(16):
+        with self.assertNumQueries(14):
```
rowanseymour, line 421: Django now uses `ON CONFLICT DO NOTHING` when adding to m2m tables instead of querying beforehand to see which values already exist

PR #371, opened 2022-02-22
Add push contacts for rapidpro

megh-k:
The `push_contact` functionality was necessary for the Open Case button functionality. For this to work a few things are necessary, as summarised below:
- URNs need to be synced to CasePro to check if the URN already exists in both CasePro and RapidPro.
- URNs also needed to be hidden from the UI for privacy. I removed the `urn.path` from the frontend but left the placeholder.
- `push_contact` checks for the contact in RapidPro by the URN, if it doesn't exist, the contact is created in RapidPro and the RapidPro `uuid` is returned.
- `get_or_create_from_urn` looks for the contact in CasePro by the URN, if it does not exist, it gets the contact from RapidPro and its `uuid`, it then looks for the CasePro contact by the `uuid` to avoid duplicated, if it still doesn't exist, a new CasePro contact is made and the `uuid` is set to match the RapidPro `uuid`.
- As we make use of WhatsApp, the WhatsApp scheme is also added as a valid scheme. 
- The test were also updated to accommodate the new functionality.

### DIFF
--- a/casepro/backend/rapidpro.py
+++ b/casepro/backend/rapidpro.py
@@ -332,11 +332,11 @@ class RapidProBackend(BaseBackend):
                 msg.backend_broadcast_id = broadcast.id
                 msg.save(update_fields=("backend_broadcast_id",))
 
-    def push_contact(self, org, normalised_urn):
+    def resolve_urn(self, org, normalized_urn):
         client = self._get_client(org)
-        contact = client.get_contacts(urn=f"{normalised_urn}")
+        contact = client.get_contacts(urn=normalized_urn)
         if not contact.first():
-            contact = client.create_contact(urns=[normalised_urn])
+            contact = client.create_contact(urns=[normalized_urn])
         else:
             contact = contact.first()
         return contact.uuid
--- a/casepro/backend/rapidpro.py
+++ b/casepro/backend/rapidpro.py
@@ -335,7 +335,7 @@ class RapidProBackend(BaseBackend):
     def push_contact(self, org, normalised_urn):
         client = self._get_client(org)
         contact = client.get_contacts(urn=f"{normalised_urn}")
-        if not contact:
+        if not contact.all():
             contact = client.create_contact(urns=[normalised_urn])
         return contact.uuid
 
--- a/casepro/backend/rapidpro.py
+++ b/casepro/backend/rapidpro.py
@@ -45,6 +45,7 @@ class ContactSyncer(BaseSyncer):
             "uuid": remote.uuid,
             "name": remote.name,
             "language": remote.language,
+            "urns": remote.urns,
             "is_blocked": remote.blocked,
             "is_stopped": remote.stopped,
             "is_stub": False,
@@ -59,6 +60,7 @@ class ContactSyncer(BaseSyncer):
             or local.language != remote.language
             or local.is_blocked != remote.blocked
             or local.is_stopped != remote.stopped
+            or local.urns != remote.urns
         ):
             return True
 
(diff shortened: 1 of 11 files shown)


casepro/contacts/models.py
```diff
@@ -317,13 +318,16 @@ def get_or_create_from_urn(cls, org, urn, name=None):
...
         if not contact:
             URN.validate(normalized_urn)
+            uuid = org.get_backend().push_contact(org, normalized_urn)
+            contact = cls.objects.filter(uuid=uuid)
 
-            contact = cls.objects.create(org=org, name=name, urns=[normalized_urn], is_stub=False)
-            org.get_backend().push_contact(org, contact)
+            if not contact:
+                contact = cls.objects.create(org=org, name=name, urns=[normalized_urn], is_stub=False)
+                contact.uuid = uuid
```
rowanseymour, line 329: can't you just add `uuid=uuid` in line above?

casepro/backend/rapidpro.py
```diff
@@ -330,8 +332,14 @@ def push_outgoing(self, org, outgoing, as_broadcast=False):
                 msg.backend_broadcast_id = broadcast.id
                 msg.save(update_fields=("backend_broadcast_id",))
 
-    def push_contact(self, org, contact):
-        return
+    def push_contact(self, org, normalised_urn):
+        client = self._get_client(org)
+        contact = client.get_contacts(urn=f"{normalised_urn}")
```
rowanseymour, line 337: isn't this already a string?

casepro/backend/rapidpro.py
```diff
@@ -330,8 +332,14 @@ def push_outgoing(self, org, outgoing, as_broadcast=False):
                 msg.backend_broadcast_id = broadcast.id
                 msg.save(update_fields=("backend_broadcast_id",))
 
-    def push_contact(self, org, contact):
-        return
+    def push_contact(self, org, normalised_urn):
```
rowanseymour, line 335: `normalized_urn` for consistency

templates/cases/case_read.haml
```diff
@@ -169,7 +169,7 @@
                 .contact-field-label.col-sm-6
                   [[ urn.scheme ]]
                 .contact-field-value.col-sm-6
-                  [[ urn.path ]]
+                  [[ ]]
```
rowanseymour, line 172: At one point in time someone was using CasePro with a backend besides RapidPro and was displaying URNs. Could we maybe add a setting like `SITE_REDACT_URNS` (see other settings like `SITE_CONTACT_DISPLAY`), and then if that is set, display `••••••••` as a placeholder like we do in RapidPro?

templates/contacts/contact_read.haml
```diff
@@ -69,7 +69,7 @@
                 .contact-field-label.col-sm-6
                   [[ urn.scheme ]]
                 .contact-field-value.col-sm-6
-                  [[ urn.path ]]
+                  [[  ]]
```
rowanseymour, line 72: as above

casepro/backend/rapidpro.py
rowanseymour: let's call this method `resolve_urn` as well since it's no longer pushing a contact

rowanseymour (changes requested):

templates/cases/case_read.haml
```diff
@@ -169,7 +169,10 @@
                 .contact-field-label.col-sm-6
                   [[ urn.scheme ]]
                 .contact-field-value.col-sm-6
-                  [[ ]]
+                  - if site_redact_urns
+                    [[ "\u2022".repeat(urn.path.length) ]]
```
rowanseymour, line 173: this should be fixed length of 8 same as rapidpro - some URNs are really long and we don't want to leak even the length of the URN

rowanseymour (changes requested):

rowanseymour (approved):

